PR #2, opened 2019-05-03
Adding brem_49 and acgaunt - Not fully working or tested yet 

hayesla:
Adding these here just as a PR to get me started - I'll fix, test it and prettify

### DIFF
--- a/sunxspex/acgaunt.py
+++ b/sunxspex/acgaunt.py
@@ -1,0 +1,361 @@
+import numpy as np 
+
+def acgaunt(wave, te_6):
+    """
+    Calculate continuum gaunt factor using approximations
+    of R. Mewe (18-JUN-85) to full calculations of 
+    paper VI (Arnaut and Rothenflug for ion balances).
+    
+
+    Parameters
+    ----------
+    wave : wavelength in angstrom (1-d vector or scaler)
+    te_6 : temperature in 10^6 K (1-d vector or scaler)
+
+    Returns
+    -------
+    cgaunt(n_elements(te_6), n_elements(wave))
+    array of approximate continuum gaunt factors
+
+    Optional Outputs
+    ----------------
+    G1 = free-free Gaunt factor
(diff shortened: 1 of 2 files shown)


sunxspex/brem_49.py
```diff
@@ -0,0 +1,34 @@
+import numpy as np 
+import acgaunt
+
+def brem_49(kt, E):
```
DanRyanIrish: FYI: https://github.com/DanRyanIrish/sunxspex/blob/master/sunxspex/models.py#L9

hayesla:
Oh yes! This is just a place holder at the moment for myself

hayesla:
I'll fix that over weekend

sunxspex/acgaunt.py
```diff
@@ -0,0 +1,361 @@
+import numpy as np 
+
+def acgaunt(wave, te_6):
```
DanRyanIrish: Perhaps this should be a class where the data currently hard coded in the function is in a (or a few) files.  On instantiation of the class the files can be read once and then the function can be a class method which has access to the data without having to reread it every time.

DanRyanIrish:
Thanks @hayesla!  Great start!

DanRyanIrish:
Don't let it interfere with your weekend, or higher priorities.

Although it is nice I'm not the only one with a PR!

sunxspex/acgaunt.py
hayesla: Yep _definitely_ agree, at the moment it's literally just a direct copy of IDL --> Python 😋

DanRyanIrish: Great!  We're on the same page!

DanRyanIrish:
Hi @hayesla.  What's the status of this PR?  Are there plans to work further on it?

hayesla:
Hi @DanRyanIrish  - I'll be back on it today and will update

PR #9, opened 2019-05-17
WIP: Chianti_kev_lines 2

DanRyanIrish:
This PR further develops the ```chianti_kev``` functions and gets the output ```chianti_kev_lines``` for a single temperature agreeing with the output of the SSW routine.


DanRyanIrish:
Currently there is still a factor of 2 discrepancy.  That must be fixed before merging.

Issue #21, opened 2019-06-20
Add code to calculate bremsstrahlung emission given electron distribution


samaloney:
So I'm planning on having a go at a very basic python version of the bremsstrahlung code. After talking with @rschwartz70 it seems like good point for separation of concerns
```python
electron_dist = some_function_generation_electron_dist(electron_paramers) # e.g. f_pow
photon_dist = calculate_bremsstrahlung(electron_dist, bremsstrahlung_parameters)
```
so the bremsstrahlung integration is separated from the generation of the electron distribution.

PR #23, opened 2019-11-20
brm2_thicktarget port

samaloney:
Port SSW IDL code as is to python as a starting point to #21

There are some very obvious and some not so obvious improvements that can be made, units for example. But I think without some upfront design even the improved code would ultimate suffer from the same limitations so I don't know if this would be time well spent or not. The main issue is the spread of different logical functions mixed into single implented functions or procedures. I thinks the most beneficial route would be to try and separate these logical functions. 

At a high level the task is given a array of photon energies and an electron distribution, low, high electron energy integration limits or cutoffs and a model of the electron plasma interaction calculate the photon flux at the given photon energies.

So looking at the current code and general task there seems to be three major components that can be generalised and abstracted to some degree:
* generating electron flux distributions or integrals (e.g. power law, broken power law, kappa, ...)
* model the electron plasma interaction (thin, thick, ....?)
* integrating the election distribution and model from the current photon energy to high electron energy for each photon energy
    * current integration code can probably be replaced with scipy quadrature and or fixed_quad but ideally should leave open to other integration schemes

We may want to keep this code as is or at least a cleaned up version for reference and comparison to what we eventually end with?


rschwartz70:
I was looking at the roadmap and it suggested that we wait for Richard(me) to put my changes to the non-thermal brem code online.  I wouldn't wait for me to get to that. I have pressing RHESSI issues and I don't think it serves a purpose. Both Shane and Danny understand the primary code issues for the numpy version and I forget who is ahead on coding them. It's perfectly fine to build the new code without updating the old as the newer version won't produce different results. Maybe I can't find them here but the python non-thermal codes should be here in developmental branches even at this fetal stage well past embryonic.  And none of the IDL code came down from Mt Sinai with Moshe.

samaloney:
Yea so tox lets us test against different versions of python so python3.6  (py36) python3.7 (py37).

rschwartz70:
When is our next tag up, meet up, telecon? Is there a place where this is
posted?

sunxspex/tests/test_brem.py
```diff
@@ -0,0 +1,66 @@
...
+from sunxspex import emission
+
+
+def test_brem_collisional_loss():
+    photon_energies = np.array([1.0, 10.0, 100.0, 1000.0])
+    res = emission.collisional_loss(photon_energies)
+    # IDL code to generate values taken from dedt variable
+    # Brm_ELoss, [1.0, 10.0, 100.0, 1000.0], dedt
+    # print, dedt, format='(e0.116)'
+    assert np.array_equal(res, [3.6275000635609979e+02, 1.2802678240553834e+02,
```
rschwartz70, line 13: Do these constants force double precision arithmetic when our knowledge doesn't begin to punch below a percent?  I'm not a fan of unnecessary DP arithmetic for computations that don't need it and which would slow down repeated computations.

PR #34, opened 2020-03-19
Use sunpy remote data for ssw files

### LINKED ISSUE #8: Local file paths currently hardcoded
The path to the ```chianti_lines_1_10_v71.sav``` file is currently hardcoded as the location on my machine.  This needs to be treated more intelligently.  Perhaps it should live in a data directory or be downloaded and stored via sunpy's database.

### LINKED ISSUE #22: Download necessary chianti files directly remove to remove dependancy on local ssw install.
Now that SSW updates have to be over http the entire SSW tree is web accessible so we can just download the chianti files from [here](https://hesperia.gsfc.nasa.gov/ssw/packages/xray/dbase/chianti). If we wanted to get fancy we could check for a local ssw versions first.

Cadair:
You might have opinions on the names I used for the files.

closes #8 and fixes #22


sunxspex/io.py
```diff
@@ -187,58 +167,47 @@ def chianti_kev_line_common_load(linefile=None):
...
+            "edge_str": edge_str,
+            "ctemp": contents["ctemp"],
+            "chianti_doc": _clean_chianti_doc(contents["chianti_doc"])
+                            }
 
     return zindex, continuum_properties
 
 
-def load_xray_abundances(abundance_type=None, xray_abundance_file=None):
+@manager.require('xray_abun',
```
samaloney:
```suggestion
@manager.require('xray_ abundances,
```

sunxspex/thermal_spectrum.py
```diff
@@ -10,18 +10,14 @@
...
-
-class ChiantiThermalSpectrum():
+class ChiantiThermalSpectrum:
     """
     Class for evaluating solar X-ray thermal spectrum using CHIANTI data.
     """
     @u.quantity_input(energy_edges=u.keV)
-    def __init__(self, energy_edges, linefile=FILE_IN,
-                 xray_abundance_file=None, abundance_type=None,
+    def __init__(self, energy_edges,
```
samaloney: There might be cases where we want to use non-default files @DanRyanIrish @hayesla ?

Cadair: You can override files using the [`manager.override_file`](https://docs.sunpy.org/en/latest/api/sunpy.data.data_manager.DataManager.html#sunpy.data.data_manager.DataManager.override_file) context manager, so it would be something like:

```python
from sunpy.data import manager

with manager.override_file("xray_abundances", url=...):
    sunxspex.whatever()
```

sunxspex/io.py
```diff
@@ -272,9 +237,7 @@ def load_xray_abundances(abundance_type=None, xray_abundance_file=None):
     # If kwargs not set, set defaults
     if abundance_type is None:
         abundance_type = "sun_coronal"
-    if xray_abundance_file is None:
-        xray_abundance_file = os.path.expanduser(os.path.join(SSWDB_XRAY_CHIANTI,
-                                                     "xray_abun_file.genx"))
+    xray_abundance_file = manager.get("xray_abun")
```
Cadair:
```suggestion
    xray_abundance_file = manager.get("xray_abundances")
```

sunxspex/thermal_spectrum.py
hayesla: I like this!

sunxspex/io.py
```diff
@@ -8,23 +8,16 @@
...
 from sunpy.io.special.genx import read_genx
 from sunpy.time import parse_time
+from sunpy.data import manager
 
-SSWDB_XRAY_CHIANTI = os.path.expanduser(os.path.join("~", "ssw", "packages",
-                                                     "xray", "dbase", "chianti"))
 
-
-def chianti_kev_line_common_load_light(linefile=None):
+def chianti_kev_line_common_load_light():
```
DanRyanIrish: If the user has there own local chianti_lines file, can they still read it somehow using this function?

sunxspex/thermal_spectrum.py
DanRyanIrish: Ah, I see this is the same comment I made above.  I agree with @samaloney that it's very possible users will want to use their own local versions of this file.  The solution @Cadair suggests looks elegant.  But I don't think it's intuitive since most of the X-ray community is not very familiar with Python.  So we need to document how to do this somewhere.  At least in a docstring here.  Probably also in a guide/tutorial at a later date.

hayesla: for sure - I think an example of this would be good. I guess most users wouldn't have known you could do this in the IDL version anyway :P 

But yes examples all the way 💯

sunxspex/io.py
hayesla: I think it would be using the same as above with `sunpy.data.manager` no?

PR #37, opened 2020-03-20
adding docstring about how to overwrite default CHIANTI file

hayesla:
This PR is adding a docstring on how to overwrite the default file within the sunpy remote data manager.

### DIFF
--- a/sunxspex/thermal_spectrum.py
+++ b/sunxspex/thermal_spectrum.py
@@ -54,13 +54,12 @@ class ChiantiThermalSpectrum:
         The default CHIANTI data used here is contained within the `sunpy.data.manager` and is collected
         from `https://hesperia.gsfc.nasa.gov/ssw/packages/xray/dbase/chianti/chianti_lines_1_10_v71.sav`.
         If the user would like to use a different file, the default can be overwritten using the context 
-        manager `sunpy.data.manager`. For example:
+        manager `sunpy.data.manager`. 
+        For example:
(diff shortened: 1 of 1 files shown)


DanRyanIrish:
Does the file have to be remote?  How does this work if your file is local?

sunxspex/thermal_spectrum.py
```diff
@@ -49,6 +49,21 @@ def __init__(self, energy_edges,
...
+        If the user would like to use a different file, the default can be overwritten using the context 
+        manager `sunpy.data.manager`. 
+        For example:
+        >>> import numpy as np
+        >>> from astropy import units as u
+        >>> from sunxspex import thermal_spectrum
+        >>> from sunpy.data import manager
+        >>> energy_bins = np.arange(3, 100, 0.5)*u.keV
+        >>> my_file = "https://hesperia.gsfc.nasa.gov/ssw/packages/xray/dbase/chianti/chianti_lines_1_10_v70.sav"
+        >>> with manager.override_file("chianti_lines_1_10", uri=my_file):
```
DanRyanIrish: I think also you should skip this in the doc tests.  Add ```# doctest: +SKIP``` to the end of this line and the next.

hayesla:
> Does the file have to be remote? How does this work if your file is local?

I think it may have to be remote from my reading of the `manager.override_file()` (@Cadair any thoughts?)

There should be a way though, I'll look into it more.

PR #58, opened 2022-04-05
Integrals

elastufka:
## PR Description

Code that performs thick/thin target integration is shorter, more legible, slightly more Pythonic, and gives the same results as the IDL code (see tests)

Unfortunately, it isn't any faster but hopefully it is an easier starting point for improvement than the IDL translation

## PR Checklist

- [X] I have followed the guidelines in the [Contributing document](https://docs.sunpy.org/en/latest/dev_guide/contents/newcomers.html)
- [X] Changes follow the coding style of this project
- [X] Changes have been formatted and linted
- [ ] Changes pass pytest style unit tests (and `pytest` passes).
- [] Changes include any required corresponding changes to the documentation
- [ ] Changes have been rebased and squashed to the minimal number (typically 1) of relevant commits
- [X] Changes have a descriptive commit message with a short title
- [ ] I have added a `Fixes #XXXX -` or `Closes #XXXX -` comment to auto-close the issue that your PR addresses

Addresses #51 but does not fix or close it

PR #64, opened 2022-04-19
Feat refactor integration

samaloney:
## PR Description

This PR aims to refactor the current integration functions to be more modular and support custom integration routines in the future. This will be split into the following three steps:

1.   - [x] Refactor IDL base code to similar API as `scipy.integrate.fixed_quad` also add extension of `fixed_quad` to handle vector limits
2.  - [ ] Refactor bremstaling code to use newly refactored 
3.  - [ ] If results match switch to scipy based integration and remove IDL based


samaloney:
@elastufka this is still in a work in progress but since you've been working on this too any early feedback would be great

sunxspex/integrate.py
```diff
@@ -0,0 +1,192 @@
+import numpy as np
+from scipy.integrate._quadrature import _cached_roots_legendre
+from scipy.special import lpmv
+
+
+def _legendre_roots(a, b, npoints, eps=3e-14):
```
elastufka, line 8: Is there a reason to keep all this? points and weights can be obtained straight from numpy and the interval transformed from [-1,1] to [a,b] easily, eg:

```
xi,wi=np.polynomial.legendre.leggauss(N)
xm=0.5*(b+a)
xl=0.5*(b-a)
new_xi=np.repeat(xm,n).reshape(xi.shape[0],n)+np.outer(xl,xi)
new_wi=np.outer(wi,xl).T
```

sunxspex/integrate.py
```diff
@@ -0,0 +1,192 @@
+import numpy as np
+from scipy.integrate._quadrature import _cached_roots_legendre
+from scipy.special import lpmv
+
```
elastufka, line 4: could consider using [autoray](https://github.com/jcmgray/autoray) for true freedom of backend. Probably only useful if the fitting infrastructure uses it too, so that tensorflow, torch, jax etc. could be used to speed that up if you've got a GPU on hand.

sunxspex/emission.py
```diff
@@ -578,14 +541,14 @@ def split_and_integrate(*, model, photon_energies, maxfcn, rerr, eelow, eebrk, e
     # Part 1, below en_val[0] (usually eelow)
     if model == 'thick-target':
         if P1.size > 0:
-            print('Part1')
+            logging.debug('Part1')
```
elastufka, line 548: is there any argument besides legacy for keeping the code in 3 very similar parts like this?

samaloney: No just not in scope of for this PR

sunxspex/integrate.py
samaloney: I guess my plan was to remove the `gauss_legendre` method and only use the `fixed_quad` one if the results are the same which they seem to be or maybe keep it for backward compatibility and use `fixed_quad` as default

sunxspex/emission.py
elastufka: okay. so is that also why nothing has been done about #58 ?

sunxspex/emission.py
```diff
@@ -353,67 +360,64 @@ def get_integrand(*, model, electron_energy, photon_energy, eelow, eebrk, eehigh
...
     """
-    Calculate the positions and weights for a Gauss-Legendre integration scheme.
+    Return the value of the integrand for the thick- or thin-target bremsstrahlung models.
 
     Parameters
     ----------
-    x1 : `numpy.array`
-
-    x2 : `numpy.array`
+    x_log : `numpy.array'
```
DanRyanIrish, line 369:
```suggestion
    x_log : `numpy.array`
```

sunxspex/emission.py
```diff
@@ -284,8 +291,8 @@ def bremsstrahlung_cross_section(electron_energy, photon_energy, z=1.2):
     return cross_section
 
 
-def get_integrand(*, model, electron_energy, photon_energy, eelow, eebrk, eehigh, p, q, z=1.2,
-                  efd=True):
+def _get_integrand0(*, model, electron_energy, photon_energy, eelow, eebrk, eehigh, p, q, z=1.2,
```
DanRyanIrish, line 294: Is this used anywhere or has it just been renamed to keep a copy of the old `get_integrand`?

sunxspex/emission.py
DanRyanIrish, line 369: Is this just `np.log10(electron_dist)`?  If so, then perhaps it's better to just derive it from `electron_dist` rather than potentially having two inconsistent inputs.

sunxspex/emission.py
```diff
@@ -472,26 +476,50 @@ def integrate_part(*, model, photon_energies, maxfcn, rerr, eelow, eebrk, eehigh
     intsum = np.zeros_like(photon_energies, dtype=np.float64)
     ier = np.zeros_like(photon_energies)
 
+    if integrator is None:
+        integrator = gauss_legendre
+    elif not callable(integrator):
+        raise TypeError('integrator must ba a callable')
```
DanRyanIrish, line 482:
```suggestion
        raise TypeError('integrator must be a callable')
```

sunxspex/emission.py
```diff
@@ -472,26 +476,50 @@ def integrate_part(*, model, photon_energies, maxfcn, rerr, eelow, eebrk, eehigh
...
+        clight = const.get_constant('clight')
+
+        electron_energy = 10**x_log
+        brem_cross = bremsstrahlung_cross_section(electron_energy, photon_energy, z)
+        collision_loss = collisional_loss(electron_energy)
+        pc = np.sqrt(electron_energy * (electron_energy + 2.0 * mc2))
+
+        density = electron_dist.density(electron_energy)
+        if model == 'thick-target':
+            return (electron_energy * np.log(10) * density * brem_cross * pc
```
DanRyanIrish, line 501: Looks like this is the same as [line 408](https://github.com/sunpy/sunxspex/pull/64/files#diff-720f0f2daa949285d71390726972a7cb2130988f4faf4247c17d9cc76fba5052R408).  Perhaps it should be moved into it's own simple function, e.g. `_thick_target_func` so there's only one place it's defined.

Same for the thin target case.

DanRyanIrish (changes requested):
Hi @samaloney.  This looks good.  Here are a few comments.

sunxspex/emission.py
samaloney: This can also be removed now

sunxspex/emission.py
samaloney: These are the points the function is evaluated at for the the integration as it's done in log space so are points so it not the same as the `electron_dist`

PR #69, opened 2022-05-09
Add tests to io module

samaloney:
## PR Description

Add tests for `sunxspex.sunxspex_fitting.io` module some of them are probably bit much as they are really just wrappers around `fits.open`.

* Small refactor of name and docstrings
* Add to API docs


KriSun95 (approved):
Thanks for fixing these doc strings and adding tests! I get what you mean when you say some of the test might be overkill at the minute but they will definitely help in the future. For example, when the STIX functions will most likely need edited.

KriSun95 (approved):

DanRyanIrish (approved):

Issue #76, opened 2022-07-13
Some of the thin/thick target test are not actually testing the code


samaloney:
### Describe the bug

Some of the test cases are missing assert statements e.g. `np.allclose(a, b)` instead of `assert np.allclose(a, b)` when enabled some of the tests fail.

### To Reproduce

import sunpy

sunpy.map.Map(...)
etc

samaloney:
So it seems the test was never actually asserting the comparisons even when I first added it 🤦 see this commit

https://github.com/sunpy/sunxspex/commit/373d0aae312fdbce8d5eb0b807416bff5f917fdc#diff-1320ad472d160b5a648de74b8da05100fe0985f4d25b0ee220722eb3b47ab532

and code 

https://github.com/sunpy/sunxspex/blob/373d0aae312fdbce8d5eb0b807416bff5f917fdc/sunxspex/tests/test_brem.py#L124-L159

and fixing the test back in that commit it still doesn't pass

```
>       assert_allclose(res_thick, res_idl_thick, atol=0, rtol=1e-10)
E       AssertionError:
E       Not equal to tolerance rtol=1e-10, atol=0
E
E       x and y nan location mismatch:
E        x: array([2.251595e-30, 3.044370e-31, 3.728863e-34, 3.307037e-36,
E              1.212173e-37, 5.335296e-39, 1.642012e-40,          nan])
E        y: array([2.251596e-30, 3.044377e-31, 3.729762e-34, 3.324935e-36,
E              1.263982e-37, 6.961897e-39, 6.223255e-40, 1.162617e-40])
```

PR #78, opened 2022-07-15
Bugfix brem tests

samaloney:
`test_get_integrand` its no longer valid to compare to IDL as due to recent refactor the function is different to idl
`test_split_and_integrate` I check when the test was orringally commit and it never worked we should look into this so created new issue #76 

The main tests which combines all the function still match which is the main thing.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -15,6 +15,7 @@ python_requires = >=3.7
 setup_requires = setuptools_scm
 install_requires =
     sunpy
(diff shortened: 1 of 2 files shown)


samaloney:
Also had to pin `parfive < 2.0` to make the test and docs pass

KriSun95 (approved):

DanRyanIrish (approved):

PR #85, opened 2022-12-06
Cstat fix

KriSun95:
The C-stat likelihood doesn't handle bins of 0 observed counts correctly. In this case, revert to the Poissonian log-likelihood. This means at Data=0, ln(likelihood)=-model.

### DIFF
--- a/sunxspex/sunxspex_fitting/likelihoods.py
+++ b/sunxspex/sunxspex_fitting/likelihoods.py
@@ -254,7 +254,7 @@ class LogLikelihoods:
         A float, the c-stat log-likelihood (to be maximised).
         """
 
-        model_counts, observed_counts, observed_count_errors = np.array(model_counts).squeeze(), np.array(observed_counts), np.array(observed_count_errors)
+        model_counts, observed_counts, observed_count_errors = np.array(model_counts).ravel(), np.array(observed_counts), np.array(observed_count_errors)
 
         # C-stat (XSPEC) - has data term in it so the lower the -2*ln(L) number the better the fit, this is the ln(L) though
         # Best value is 0, if obs_counts>mod_counts or obs_counts<mod_counts then likelihoods<0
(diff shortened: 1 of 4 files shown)


samaloney (changes requested):
Nice fix, I think it needs a test e.g. calling `cstat_loglikelihood` function with some data with zeros and making sure it behaves as intended.

Also none of these show up in the doc build which makes it hard to review the changes to the docs would be good to get that sorted maybe not in this PR. I've open a different PR to add the docs #94 you can see a built version here https://sunxspex--94.org.readthedocs.build/en/94/

sunxspex/sunxspex_fitting/likelihoods.py
```diff
@@ -250,10 +254,18 @@ def cstat_loglikelihood(self, model_counts, observed_counts, observed_count_erro
...
         # Best value is 0, if obs_counts>mod_counts or obs_counts<mod_counts then likelihoods<0
         # Obvious since e^0=1
-        likelihoods = np.array(observed_counts) * (np.log(np.array(model_counts)/np.array(observed_counts)) + 1) - np.array(model_counts)
+        likelihoods = observed_counts * (np.log(model_counts/observed_counts) + 1) - model_counts
+
+        # if zero observed counts in a bin, then Stirling Approx. not good (only good for obs_counts>=1, if =0 then ->-inf, but Poisson->-mod_counts)
+        # defer to Poisson for the nans in this case
+        _zero_data = (observed_counts == 0) & (model_counts > 0)
+        if len(observed_counts[_zero_data]) > 0:
+            likelihoods[_zero_data] = -model_counts[_zero_data]
```
samaloney, line 268: Is the if statement necessary? `_zero_data` is a boolean mask so should always work. Unless it a perforce thing but might be faster too?

samaloney (approved):

PR #87, opened 2022-12-06
count_rate2counts should be dealing with the average, so no need for integers

KriSun95:
Not right to cast model counts to integers since this is supposed to be the model average in the time range being looked at and so it can be a float, it is the count data that must be integers. Casting to integers the way I did can also cause weird value overflows.

### DIFF
--- a/changelog/87.bugfix.rst
+++ b/changelog/87.bugfix.rst
@@ -1,0 +1,1 @@
+Values are now left as floats when converting the model count rates (counts/second) to just counts since the model represents the average number of counts; therefore, there is not need for these to be integers.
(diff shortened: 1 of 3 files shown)


changelog/87.bugfix.rst
```diff
@@ -0,0 +1 @@
+Values are now left as floats when converting the model count rates (counts/second) to just counts since the model represents the average number of counts; therefore, there is not need for these to be integers.
```
samaloney, line 1:
```suggestion
Removed cast to `int` when converting the model count rates (counts/second) to just counts since the model represents the average number of counts, there is not need for these to be integers.
```

samaloney (approved):
Looks good thanks. Small suggestion for the change log you can take it or leave it.

samaloney (approved):

PR #89, opened 2022-12-06
Fixed the way random walker starts are picked over a boundary 

KriSun95:
The way I did this in the first place was stupid. Originally the parameter space bounds were multiplied by 100 and then I choose random integer values (why do I want to make everything an integer!?) within the new scaled bounds. This is obviously not right.

The new way is done as L+(U-L)*R, where L and R are the lower and upper bounds for the parameter space and R is a random value between 0 and 1. This should now give values that make sense no matter the bounds.

### DIFF
--- a/changelog/89.bugfix.rst
+++ b/changelog/89.bugfix.rst
@@ -1,1 +1,1 @@
-Starting random walker positions over the parameter space for MCMC analysis are now obtained correctly and does not rely on all upper parameter bounds being >0.01.
+Fix bug where MCMC random walker starter positions were not correctly calculated.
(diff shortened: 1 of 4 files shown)


changelog/89.bugfix.rst
```diff
@@ -0,0 +1 @@
+Starting random walker positions over the parameter space for MCMC analysis are now obtained correctly and does not rely on all upper parameter bounds being >0.01.
```
samaloney, line 1:
```suggestion
Fix bug where MCMC random walker starter positions were not correctly calculated.
```

samaloney (changes requested):
Thanks, nice catch. I think a test calling `_boundary_spread` with some vales and comparing to expected would be really nice.

settwi:
I think this approach still assumes that the parameter bounds are close to the best value.

For things like emission measure I've often made the bounds very large (like 1e-5 to 1e20). The `'mag_order'` approach makes more sense for starting close to the minimizer parameters.

Why not make `'mag_order'` the default?

PR #90, opened 2022-12-07
Change chanti data urls

samaloney:
## PR Description

Update Chanti .sav url in test to point to lmsal server rather than hesperia


samaloney:
Yea I agree, ideally not depend on internet server for test or at least use all the possible source lmsal, nasasoho and hesperia.

Also I just changed back to the main OpenAstronomy template repo as the PR I needed was merged there.

Issue #97, opened 2023-05-24
STIX file loading and lightcurve/spectrogram plotting


natsat0919:
### Describe the bug

The STIX spectral and srm file loading functions don't support the updated version of STIX files, these are mainly formatting issues (e.g. change of keyword in fits file from DATE-BEG to DATE_BEG), but also other changes such as change of time units from seconds to deci-second. 

These issues (even when fixed for loading in new files) then result in plotting errors. The highlighted background and event time regions in STIX light curves and spectrograms are shifted, therefore the plots are showing incorrect regions than what the user selected.

### Screenshots

In this example I was loading in STIX observation and was selecting:

Background region: from "2020-06-07T21:30:00.000" to "2020-06-07T21:40:00.000"

Event region: from  "2020-06-07T21:40:00" to "2020-06-07T21:43:45")

You can see that in the light curve example, the highlighted time regions are all shifted. In the spectrogram example, the highlighted regions are overlapping.

![lc](https://github.com/sunpy/sunxspex/assets/61255393/57e8b0b4-9f5b-4e5a-b298-50396f7d0db7)
ping from about 21:39 to 21:41.
![sg](https://github.com/sunpy/sunxspex/assets/61255393/813f1f79-65ce-42b3-baf5-e7ce3928bc52)

PR #98, opened 2023-05-24
Fixes STIX file loading and lightcurve/spectrogram plotting

natsat0919:
This PR fixes the functions for handling STIX spectral and SRM files, this means that the Sunxspex class will be apple to correctly load in the new/final version of STIX files. I'm mainly fixing formatting of keywords and data, change of time units, format of energy bins and counts that are given in the spectral files. I have also fixed the incorrect plotting of background and event time-ranges for STIX lightcurves and spectrograms (examples in issue #97).

### DIFF
--- a/sunxspex/sunxspex_fitting/stix_spec_code.py
+++ b/sunxspex/sunxspex_fitting/stix_spec_code.py
@@ -28,8 +28,8 @@ def _get_spec_file_info(spec_file):
     """
     sdict = io._read_stix_spec_file(spec_file)
 
-    times_mids = sdict["2"][1]["time"]  # mid-times of spectra, entries -> times. Mid-times from start of observation
-    time_deltas = sdict["2"][1]["timedel"]  # times deltas of spectra, entries -> times
+    times_mids = _ds_times2s_times(sdict["2"][1]["time"])   # mid-times of spectra, entries -> times. Mid-times from start of observation
+    time_deltas = _ds_times2s_times(sdict["2"][1]["timedel"]) # times deltas of spectra, entries -> times
     time_diff_so2e = sdict["0"][0]["EAR_TDEL"]  # time difference between Sun2Earth and Sun2SO, so time at earth for measurement is this difference added on to the actual detection time
     # spectrum number in the file, entries -> times # spec_num = sdict["1"][1]["SPEC_NUM"]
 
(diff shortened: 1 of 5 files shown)


DanRyanIrish:
Hi @natsat0919.  Thanks so much for contributing this PR.  Your changes appear to be breaking a lot of the tests.  Can you clarify why this is?  Is it possible you've made changes in certain places that a valid for STIX but not other instruments?

natsat0919:
Hi @DanRyanIrish, I'm missing a changelog file that I'll add. I'm still trying to figure out why the other tests are failing, but I think that it's because of changing the fundamental loading function for STIX files and the tests aren't adjusted to that? The changes that I have done were only for STIX (mainly in the stix_spec_code.py), so it shouldn't affect any other instruments

settwi:
Following up on our discussion @natsat0919 and @KriSun95: I think we should be careful and explicit as to which kind of "spectrogram" we load in, i.e. the one produced by IDL or the one from the data center, and pay particular attention to the statistical errors that are being used. I would be happy to review some of the code in this PR and see how we can make sure that distinction is clear. Or we can merge the PR and do that later. Thoughts?

natsat0919:
@settwi I agree, we definitely need to get that sorted as a priority. I think right now is the best to get this out as the basic working version (just updating what we already have) and build up from there? Also the STIX srm and spectral files tests are outdated (hence I got so many fails) so I'm sorting that out as well. Meanwhile, we can think what's the best way to implement the "distinction" when loading in?

KriSun95:
This looks great, thanks @natsat0919 ! I've had a play about with the changes and it works fine for me. Playing about with it has pointed out a small but important bug elsewhere (that I need to fixed) but I can't see anything wrong with this PR. I definitely agree with both you and @settwi about moving on to think about how to separate loading in of the two STIX file formats for the spectra (data centre and IDL).

I'll return to this later today, add a review for the PR, and then move on to approve the merge if there are no other objections.

KriSun95 (approved):
This looks great and gives us a good foundation to actually start working with/support STIX data. We can use this contribution to test any future changes using real data and use it as a good starting point.

DanRyanIrish:
Do we consider this PR ready to merge?

natsat0919:
@DanRyanIrish yes, I just did some very minor refactoring changes after Kris's approval. It should be ready to merge

sunxspex/sunxspex_fitting/stix_spec_code.py
```diff
@@ -40,8 +40,8 @@ def _get_spec_file_info(spec_file):
     _plus_half_bin_width = np.ceil(time_deltas/2)
     t_hi = times_mids + _plus_half_bin_width
 
-    spec_stimes = [Time(sdict["0"][0]["DATE_BEG"], format='isot', scale='utc')+TimeDelta(time_diff_so2e * u.s)+TimeDelta(dt * u.ds) for dt in t_lo]
-    spec_etimes = [Time(sdict["0"][0]["DATE_BEG"], format='isot', scale='utc')+TimeDelta(time_diff_so2e * u.s)+TimeDelta(dt * u.ds) for dt in t_hi]
+    spec_stimes = [Time(sdict["0"][0]["DATE-BEG"], format='isot', scale='utc')+TimeDelta(time_diff_so2e * u.s)+TimeDelta(dt * u.s) for dt in t_lo]
+    spec_etimes = [Time(sdict["0"][0]["DATE-BEG"], format='isot', scale='utc')+TimeDelta(time_diff_so2e * u.s)+TimeDelta(dt * u.s) for dt in t_hi]
```
samaloney, line 44: I'm not sure this change is needed if the goal is to get the value in seconds `to_value(u.s)` could be used and would remove the somewhat arbitrary `* (10**(-2))` below.

samaloney (changes requested):

sunxspex/sunxspex_fitting/stix_spec_code.py
KriSun95: Good catch! The `_ds_times2s_times`  function could be removed altogether, this seems like a relic from when I was coding for multiple file structures. Would it be better to just convert `times_mids` and `time_deltas` (lines 31 and 32, respectively) on the same lines and remove the `_ds_times2s_times` function? Say change line 31 to something like `time_mids = (sdict["2"][1]["time"]<<u.ds).to_value(u.s)` and change line 32 in a similar way?

natsat0919: Oh yeah, that's a good point. Thanks for raising that change. I have just corrected that (the time is in units of cs) by multiplying tds by u.cs and keeping time_mids and time_deltas without units (assigning units to the arrays breaks the code if you try to select an event/background time region).

samaloney (approved):

Issue #107, opened 2023-06-27
Unable to pass instrument loader classes to the main fitting class directly


KriSun95:
### Describe the bug

# Bug

The main fitting class [`sunxspex.sunxspex_fitting.fitter.SunXspex`](https://github.com/sunpy/sunxspex/blob/master/sunxspex/sunxspex_fitting/fitter.py#L62) cannot take an instrument loader class directly because of a [bug](https://github.com/sunpy/sunxspex/blob/master/sunxspex/sunxspex_fitting/data_loader.py#L143) in the [`sunxspex.sunxspex_fitting.data_loader.LoadSpec`](https://github.com/sunpy/sunxspex/blob/master/sunxspex/sunxspex_fitting/data_loader.py#L20).

The bug causes an error when passing a user made instrument loader since I've written to check the user class is a subclass, instead of just an instance, of the [`instruments.InstrumentBlueprint`](https://github.com/sunpy/sunxspex/blob/master/sunxspex/sunxspex_fitting/instruments.py#L30) class. The current bug makes it difficult to troubleshoot current instrument loaders and create new ones.

The fitting class *should* be able to take the instrument loaders and future updates/discussions might make this the preferred way to pass information to the fitting class. I.e., rather than passing different instrument data files to the fitting class directly the files should be handed to the relevant instrument loader first then the instance of the loader passed to the fitting class.

# Proposed Solution

Although more work will be done on the loading side of this, fixing this seems to be a good thing to do. I will open a PR with my solution of replacing the incorrect line with one that only checks if the instrument class being passed is an *instance* of `instruments.InstrumentBlueprint`. Tests will also be written to ensure expected behaviour.

### To Reproduce

```
from sunxspex.sunxspex_fitting.instruments import NustarLoader
from sunxspex.sunxspex_fitting.fitter import SunXspex

# could be any original instrument class, not just NustarLoader()
class nu(NustarLoader):
    def __init__(self, *args, **kwargs):
        NustarLoader.__init__(self,*args, **kwargs)

# give my new instrument loader my specific data
d_dir = "/Users/kris/Documents/umnPostdoc/projects/analysis/nustarNov2021/nestedSampling/analysis/ns_products/"
nu_inh = nu(pha_file=d_dir+'evt_2112_6to10keV/nu20618003001A06_cl_grade0_sr.pha')

# pass the new instrument loader instance to the main fitting class
f = SunXspex(nu_inh)
```

### Screenshots

[...]

PR #108, opened 2023-06-27
Pass instrument loader directly to fitter class

KriSun95:
## PR Description

As described in #107, the main fitting class doesn't support being handed an instrument loader class because of a bug. It is probably cleaner to hand instrument data files to a specific loader, *do stuff*, then pass the instance of the loader to the main fitting class rather than crudely just passing the data files directly to the fitting class. This PR aims to fix this bug.

The instrument class instance should now be able to be passed to the fitting class without error now, it doesn't impact any of the other code that's written, and has tests to avoid this bug being reintroduced.

I'd be happy to hear comments on the fix if people have them. This part of the code will be gradually tidied up and rearranged over time but I think having this bug fixed is worthwhile since it will help troubleshoot existing instrument loaders and help create new ones.

### DIFF
--- a/changelog/108.bugfix.rst
+++ b/changelog/108.bugfix.rst
@@ -1,0 +1,1 @@
+Fixed bug where passing a user defined instrument loader class to :class:`sunxspex/sunxspex_fitting/fitter.SunXspex` caused an error.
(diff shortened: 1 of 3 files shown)


settwi:
i think in the future we should drop the file input options. just commenting here for visibility...

KriSun95:
> i think in the future we should drop the file input options. just commenting here for visibility...

I hint at this in #107, as I do in the PR description, but it isn't the clearest so thanks for being explicit here. I'll add a comment to the Issue with the same sentiment.

samaloney (approved):

PR #110, opened 2023-07-10
STIX data center file loading and error propagation

### LINKED ISSUE #109: Adjust STIX file loading to include more errors
### Provide a general description of the issue or problem.

STIX has a lot of errors to consider because it compresses counts and triggers and the gain calibration shifts around.

We should include those for spectroscopy.

I'll attach a PR shortly

settwi:
## PR Description

Adds more error propagation for STIX files and reformats some of how the RHESSI/STIX loader classes deal with returning many arrays.

TBH I think we should decouple RHESSI and STIX loaders as much as possible. They are similar in some ways but STIX has pixel data which we will need to handle totally separately.

Fixes #109 

I need to run the tests on the updated files still

WS


settwi:
Pending the style checks I think this PR is ready to be undrafted so I'll mark it as such.

Here is the example notebook: https://github.com/settwi/sunxspex/blob/stix-error-propagation/examples/stix/stix-example.ipynb

sunxspex/sunxspex_fitting/stix_spec_code.py
```diff
@@ -117,20 +121,105 @@ def _get_srm_file_info(srm_file):
...
+
+    live_ratio = np.exp(-BETA * nin * ASIC_TAU * tau_conv_const)
+    live_ratio /= (1 + nin * TRIG_TAU)
+    corrected_rate = count_rate / live_ratio[:, None]
+    return {
+        'corrected_rate': corrected_rate,
+        'count_rate': count_rate,
+        'photons_in': photons_in,
+        'live_ratio': live_ratio
+    }
```
samaloney, line 189: I'm not sure this kind of instrument specific code belongs in sunxspex

sunxspex/sunxspex_fitting/stix_spec_code.py
```diff
@@ -117,20 +121,105 @@ def _get_srm_file_info(srm_file):
...
+    num_dets: int,
+    counts: u.ct,
+    time_bins: u.s
+):
+    '''
+    Estimate livetime and its error from triggers and counts.
+    '''
+
+    lt_high = estimate_livetime(
+        triggers=triggers + trig_comp_err,
```
samaloney, line 205: Should these added in quadrature?

sunxspex/sunxspex_fitting/stix_spec_code.py
```diff
@@ -117,20 +121,105 @@ def _get_srm_file_info(srm_file):
...
+    )
+
+    lt_low = estimate_livetime(
+        triggers=triggers - trig_comp_err,
+        counts_arr=counts,
+        time_bins=time_bins,
+        num_detectors=num_dets
+    )
+
+    lt_err = np.abs(lt_low['live_ratio'] - lt_high['live_ratio']) / 2
```
samaloney, line 218: I need to check but I don't think the compression error are symmetric and does this change this calculation?

sunxspex/sunxspex_fitting/stix_spec_code.py
```diff
@@ -2,113 +2,117 @@
...
+    with fits.open(spec_file) as sf:
+        # get units from the fits columns rather than hard-coding
+        slice_unit = lambda hdu, c: u.Unit(sf[hdu].columns[c].unit)
+
+        time_mids = sf['data'].data['time'] << slice_unit('data', 'time')
+        dt = sf['data'].data['timedel'] << slice_unit('data', 'timedel')
+        # if odd `t` in seconds then edges are [midt-floor(bin_width/2), midt+ceil(bin_width/2)]
+        # e.g., mid_t=19, del_t=13 then edges would be [19-6, 19+7]=[13,26]
+        t_lo = time_mids - (np.floor(dt.value / 2) << dt.unit)
+        t_hi = time_mids + (np.ceil(dt.value / 2) << dt.unit)
```
samaloney, line 40: Not sure the floor / ceil are correct, take a time of 50, and tiemdel of 13 the edges should be `50 - 13/2`, `50 + 13/2` so 43.5, 56.5 not 50 - 6, and 50+7 ?

sunxspex/sunxspex_fitting/instruments.py
```diff
@@ -1522,11 +1488,58 @@ def _getsrm(self, f_srm):
...
+        # trim counts bins to SRM counts bins
+        # (some aren't computed for the SRM)
+        self.clip_energy_edges(ret)
+        return ret
+
+    def clip_energy_edges(self, srm_info: SrmFileInfo):
+        '''STIX SRM bin edges don't always line up with the spectrum edges.
+           Clip them so that matrix multiplication works.
+        '''
+        if srm_info.count_bin_edges.size > self.spec_info.count_bins.size:
```
samaloney, line 1503: We will very shortly if be moving to a much finer grid in the DRM which will never match the counts bins.

examples/stix/example-fit.png
samaloney, line 1: Not ideal to have binary data in the repo

samaloney (changes requested):
Really nice addition.

Two main concerns some of this very instrument specific code probably doesn't belong in sunxspex and related to this is the now 3 (python packages) which are have very similar but different livetime correction codes.

settwi:
I agree. Can we move some of the very instrument-specific code to stixpy? I took the correction function directly from stixdcpy but modified the time constants with updated values I got from Hannah Collier.

Thanks for the review!
Best
William

sunxspex/sunxspex_fitting/stix_spec_code.py
KriSun95: This was the case about over a year ago when I wrote the code for some of the first STIX files I got to load into Sunxspex. It's because the time bins are contiguous and the file only gave the "mid-time" and the bin-width, both given as integers of centi-seconds, so the bin edges couldn't be at 9.5 (or anything.5) unless the first time bin started at 0.5 s or all `timedels` are odd values. It definitely needs check for the files now just to be safe since `time_mids` might not be given as integers anymore.

sunxspex/sunxspex_fitting/stix_spec_code.py
KriSun95: Good point! It would be worth discussing this at the next meeting at what point we should/have to stop with the instrument specific/really specific stuff .

Issue #112, opened 2023-09-11
High-energy break of `thick_fn` should not be 150 keV


settwi:
#### [See here](https://github.com/sunpy/sunxspex/blob/81ba5ab57b565766682fb359eaadf466f33be522/sunxspex/sunxspex_fitting/photon_models_for_fitting.py#L94)

The high-energy break (?) of the thick target single power law emission function is locked to 150 keV. for RHESSI flares where the emission extends beyond 150 keV the break should be variable.

Proposed fix: make it so the break energy takes the max of the given energy array and multiplies it by 10, or something like that.

settwi:
Does the thick target function `thick_fn` have a test associated with it?
I am gonna link a PR. Seems like there aren't any official tests associated with `thick_fn`, unless I'm missing something.

PR #113, opened 2023-09-12
Fix thick target

settwi:
Change the upper bound of `thick_fn` to be energy-vector dependent instead of a fixed 150 keV.
Addresses #112


settwi:
Oops

Issue #116, opened 2023-09-18
RHESSI attenuator choice in SRM


settwi:
### Describe the feature

RHESSI has different SRMs for different attenuator states. The easiest and arguably safest thing to do is just let the user pick the attenuator state when loading in RHESSI data.

settwi:
Fixed once #143 is merged.

PR #117, opened 2023-09-18
Rhessi srm editable

settwi:
Addresses #116 by modifying `RhessiLoader` with extra `kwarg`.
Example in the `examples` folder too.


settwi:
@KriSun95, @ianan, @LinErinG --
i updated this to auto-pick the RHESSI SRM in the specified spectroscopy interval.
it also raises a warning if you span two intervals with an attenuator state change.

this is not in the style of the new refactor but rather a bandaid fix so that i don't have to keep struggling with a jerry-rigged API for a paper i am working on :p

- WS

settwi:
Annoying conflicts post-package rename so i am opening a new draft pr

Issue #119, opened 2023-10-02
Only instrument loading classes should be passed to the fitter, nothing else


KriSun95:
### Describe the feature

Previously, only files and dictionaries containing data to be fitted could be passed to the fitting code. This was improved in #107 where a bug was fixed and the instrument loader classes were allowed to be passed.

The proposed feature in this Issue is to **only** allow the instrument loaders to be passed. This will detangle a lot of code while making it simpler to separate and maintain code relating to different instruments right now and in the future.

### Proposed solution

Remove the ability to pass files and dictionaries to the fitting code. Instead, the files will be given to the relevant [instrument loader](https://github.com/sunpy/sunxspex/blob/master/sunxspex/sunxspex_fitting/instruments.py) which can then be passed to the fitting code. Likewise, custom dictionaries from a user's unsupported instrument or source can be given to the [`CustomLoader()`](https://github.com/sunpy/sunxspex/blob/master/sunxspex/sunxspex_fitting/instruments.py#L1532) and then passed.

This change should be consistent with and support any future re-factor to the code, especially when making it more modular.

PR #120, opened 2023-10-02
Only pass instrument loader

KriSun95:
Separate more of the instrument specific code from the more general fitting code. Remove the ability to pass files and dictionaries directly to the fitter. instead, the files (or dictionaries) can be passed to instrument specific (or custom) loaders and then passed to the fitting code.

This is to address #119 in order to support a more modular approach to the code and reduce complexity (also pointed out in #107).


settwi:
how about we implement this in "legacy"?
might make our lives nicer down the road :p

PR #124, opened 2023-10-18
fix fitter / instrument class to allow parallelization

settwi:
- Fix `__setstate__` and `__getstate__` associated with `pickle`ing for parallelizing e.g. MCMC properly.
- Also change the `function_creator` and `deconstruct_lambda` functions to remove the "dynamic source" side effect.
IMO that shouldn't be part of the function body and should just be done outside of the function call.
If we want that to be part of the call, it should be a member function (method) and not a free function.


settwi:
@KriSun95

PR #133, opened 2023-11-28
Rename sunxspex_fitting module to fitting.

DanRyanIrish:
## PR Description

As described in PR title.  Justification is that the package is no longer called `sunxspex` and there's no need for a module to carry the name of the package.


samaloney (changes requested):
I would suggest anything other than `fitter` when refactoring the current implementation it's doubtful it can be done in it one go so probably need to support current fitting (`fitter_legacy`) and new refactor simultaneously for some time. This was the reason this wasn't done in the initial rename of the package.

DanRyanIrish:
Ah, I see @samaloney.  So you suggest not merging this PR and keeping it as is.  Then we can create a new `fitting` module for the refactored fitting?

samaloney:
Essentially, but happy to merge this PR just with a different name to make it easier during the refactor e.g. support both for while

DanRyanIrish:
Perhaps it's easier for backwards compatibility to keep the name as is and then deprecate it once the new `fitting` module is ready.

KriSun95:
I like the idea of changing it to something generic like `legacy_fitter`, purely because of the past confusion in the naming. Since most of the name change has already taken place, and only just, I'm personally not as worried as much about this edit affecting backwards compatibility.

samaloney:
Is this necessary with #137?

Issue #126, opened 2023-10-23
Refactoring the instrument data handling


KriSun95:
### Provide a general description of the issue or problem.

This tackles a subset of the points raised in #81 where we define the format of the instrument data to be fitted.   

The aim is to ensure that the loading of the data and creation of the spectrum to be fitted is made modular in relation to the rest of the package. This will make the code easier to maintain and allow for easier interactions with instrumental data.

_Discussed by @KriSun95 and @settwi (23/10/2023) while focussing on the "Observational data" section of the basic flow chart in the [DASH 2023 Sunxspex poster](https://zenodo.org/records/8414868) (also shown here)._
![sunxspex_basic_flow_chart](https://github.com/sunpy/sunxspex/assets/43237137/16949eba-52a2-4312-9546-189d2f513f89)

***

## Defining basic information needed for fitting

The first point is to define what information is explicitly ***needed*** for the X-ray forward-fitting process, as well as the appropriate format and units. This basic and fundamental list includes:
* counts 
    * _1D, label:`counts`, unit:`counts`_
* counts error 
    * _1D (same length as counts), label:`counts_error`,  unit:`counts`_
* count energy bin edges 
    * _1D, label:`count_channel_bins`, unit:`keV`_
* photon energy bin edges 
    * _1D, label:`photon_channel_bins`, unit:`keV`_
* spectral response matrix 
    * _2D, label:`spectral_response_matrix`, unit:`cm^2 count photon^-1`_
* effective exposure 
    * _scalar or 1D (same length as counts), label:`effective_exposure`, unit:`s`_

An instrument loader class will contain an `event_data` attribute with this base information (as a _generic spectrum class_, perhaps) in this format which will be passed to the fitting object. Additionally, if the user wishes to include a background spectrum, a `background_data` attribute will exist and will be of the same type as `event_data` but with the background data for the event spectrum instead.

The units will be assigned to the arrays using `astropy.units` and it is be the job of the specific instrument loader to either assign or convert the data to the appropriate units from the start. Note, the units will be stripped when passing to the fitter class.

All other information (e.g., rates, mid-point energies, etc.) will be calculated from the base information, will not be changeable without changing the base information, and be contained in the wider instrument loader class if not helpful during the fitting process.

***

## Why have an instrument loader and generic spectrum class?

[...]

Issue #81, opened 2022-09-02
Refactor of Fitting Infrastructure


DanRyanIrish:
### Provide a general description of the issue or problem.

The fitting infrastructure provided by @KriSun95 is a valuable addition to the package.  To help modularity, flexibility and ease of maintenance, a refactor would be beneficial.  This issue sketches out how this could be done.

Components
* Data structure: A structure of holding the spectrum/spectra.  (Inherits from NDCube?/xarray?/Spectrum1D?)
* Photon models:  Functions providing photons as a function of energy determined by certain physical variables, e.g. temperature  (Thermal and thin/thick target models exist but need more work.)
* Spectral Response Model: Given physical variables of photon model/composition of photon models, returns count spectrum.  It is a pipeline of composed of:
  * photon model/photons models
  * instrument response including but not limited to detector response matrix
* Fitting tool: Compares observed and model counts and iterates physical variables to minimise residuals.
* These tools must be able to handle multiple spectra from multiple instruments with different spectral responses and minimise fit using the same photon model(s).

### Data Structure Requirements
* Stores and provides easy access to count data and spectral axis values.
* Handles irregular energy bin widths
* Can return energy bin centers, edges and widths.
* ...

settwi:
I like the general setup of this.
I think we should avoid inheritance unless it will offer an obvious benefit to the code. We can just use composition (i.e. a "has-a" NDCube rather than an "is-a" relationship). It makes it easier to swap things out later.

@KriSun95 have been thinking of  having a mini-workshop for a few days at some point in the near future to hash out some ideas too.

samaloney:
I think the main thing is to define the API. How that API is implemented e.g composition, inheritance or a mix is a detail, all be it an important one. Ideally we would have something something like the [NDCube SEP](https://github.com/sunpy/sunpy-SEP/blob/master/SEP-0012.md) realistically probably a simlified version 😆

DanRyanIrish:
After the discussion at the STIX meeting last week, here's an update on one option for the refactor

# Overall structure for fitting a single model to a spectrum/multiple spectra on same spectral grid
3 main objects:
* `Spectrum`: Holds the observed data.
  * Suggest this follows the API of `NDCube` and any extra methods/attributes follows the API of `specutils.Spectrum1D` (which is an instance of `NDCube`).
[...]

PR #140, opened 2024-03-01
Modular fitter attempt

settwi:
# PR Description

A crack at an updated "modular" fitting interface. Notebook included demonstrating fitting a "line" of "X-ray" data.
No real tests written yet but that's on the agenda, if we want to continue with any of this.

I am curious as to what folks think!

---

Here's a copy/paste from the "prelude" to that Jupyter notebook:

# Modular modeling framework example

This notebook shows how a fitting framework could be made,
    in the context of [issue #81](https://github.com/sunpy/sunkit-spex/issues/81) and [issue #126](https://github.com/sunpy/sunkit-spex/issues/126).

`sunkit-spex` has (or will have?) a modular interface.
First we define the data interface.
For now it is only X-ray data but we can add more data types.

Then we define the fitting interface.
A `PhotonModel` represents any kind of model which represents photons which get counted.
In the future we can add different models, like `IntegratedFluxModels` if the
    need arises; there is not a need to tie us to a base class right now.

The first example tackled here is fitting data to a line.
More complex models may be implemented however the user or developers see fit
    by modifying the `perform_fit` method in subsequent classes.
The first class implemented is a Monte Carlo $\chi^2$ fitter.

A lot of ideas here are similar to [pr #130](https://github.com/sunpy/sunkit-spex/pull/130).
I think we should pick and choose what we like from both and move from there.


settwi:
@samaloney @KriSun95 @DanRyanIrish 
tear this apart plz

PR #142, opened 2024-04-15
Add Spectrum and CountSpectrum objects

samaloney:
Heavily based off of the `Spectrum` object in specutils but also significantly stripped down. I think since the API is a subset it would easy enough to adopt the full API from specutils in the future if we decide to go that way. Since we know we will have a X-ray count spectra added a specific class for this as will have different properties.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -15,19 +15,19 @@ packages = find:
 python_requires = >=3.9
 setup_requires = setuptools_scm
 install_requires =
-    sunpy
-    parfive
-    scipy
-    xarray
-    quadpy
-    orthopy
-    ndim
-    matplotlib
-    emcee
     corner
+    emcee
+    matplotlib
+    ndcube
+    ndim
     nestle
     numdifftools
-
+    orthopy
+    parfive
+    quadpy
+    scipy
+    sunpy
+    xarray
 
 [options.extras_require]
 test =
@@ -35,6 +35,7 @@ test =
     pytest-astropy
     pytest-cov
     pytest-xdist
+
 docs =
     sphinx
     sphinx-automodapi
--- a/sunkit_spex/extern/ndcube/meta.py
+++ b/sunkit_spex/extern/ndcube/meta.py
@@ -72,6 +72,7 @@ class Meta(dict):
     axis-awareness.  If specific pieces of metadata have a known way to behave during
     rebinning, this can be handled by subclasses or mixins.
     """
+
     def __init__(self, header=None, comments=None, axes=None, data_shape=None):
         self.__ndcube_can_slice__ = True
         self.__ndcube_can_rebin__ = True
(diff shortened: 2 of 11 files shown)


samaloney:
I think the work we are doing in stixpy (TCDSolar/stixpy/pull/109) might be relevant here too @DanRyanIrish

DanRyanIrish:
I should point out a likely future direction of `NDCube` is to add an attribute contained an `NDCoords` class, which provides in easier API for accessing coordinate info.  A possible implementation can be seen in [ndcube#710](https://github.com/sunpy/ndcube/pull/710).

In order to make the Tabular WCS infrastructure work properly with edges, we would also need to finish [ndcube#713](https://github.com/sunpy/ndcube/pull/713).  

Once, these features are available, the only advantage I see for a `Spectrum` class in `sunkit-spex` would be a simpler instantiation API which doesn't require a WCS to be constructed manually beforehand.  Everything else we needed would be on `NDCube`.

Lowering the instantiation barrier for users is important.  Alternatively to a `Spectrum` class, this could be done by providing a function, `build_spectral_cube()`, which takes a data quantity, spectral axis quantity, and other optional `NDCube` inputs, and outputs an `NDCube` with the correct format.  Dummy world axes with pixel units could be assigned if the `data` array was multi-dimensional.

If we agree this is the way forward, then working on the above two `ndcube` PRs becomes the bulk of the work needed for a de-facto `Spectrum` data object.

samaloney:
> I should point out a likely future direction of `NDCube` is to add an attribute contained an `NDCoords` class, which provides in easier API for accessing coordinate info. A possible implementation can be seen in [ndcube#710](https://github.com/sunpy/ndcube/pull/710).
> 
> In order to make the Tabular WCS infrastructure work properly with edges, we would also need to finish [ndcube#713](https://github.com/sunpy/ndcube/pull/713).
> 
> Once, these features are available, the only advantage I see for a `Spectrum` class in `sunkit-spex` would be a simpler instantiation API which doesn't require a WCS to be constructed manually beforehand. Everything else we needed would be on `NDCube`.
> 
> Lowering the instantiation barrier for users is important. Alternatively to a `Spectrum` class, this could be done by providing a function, `build_spectral_cube()`, which takes a data quantity, spectral axis quantity, and other optional `NDCube` inputs, and outputs an `NDCube` with the correct format. Dummy world axes with pixel units could be assigned if the `data` array was multi-dimensional.
> 
> If we agree this is the way forward, then working on the above two `ndcube` PRs becomes the bulk of the work needed for a de-facto `Spectrum` data object.

What is the best way to get these changes, extern the whole of NDCube and add the changes from these PRs 😆

samaloney:
So I think we should maybe just merge this so we a very basic spectrum object

DanRyanIrish (approved):

KriSun95 (approved):

PR #146, opened 2024-06-05
Update requirements.txt

FelipeMFO:
Update requirements.txt


samaloney:
Hi thanks for the PR as the package is intended to be installed via pip etc we don't use a `requirements.txt`. You can install it remotely via `pip install git+https://github.com/TCDSolar/xrayvision.git` or if you cloned the repo locally `pip install -e .[all]`. (I need to check the exact syntax so if these don't work let us know)

DanRyanIrish:
@samaloney So should this PR be closed if we don't currently use requirements.txt?

samaloney:
Yea this can be closed I was just giving the author some time to respond to make sure they got sorted.

Issue #42, opened 2020-03-25
Suggestion for package structure


samaloney:
Just an idea

```
└── emission
    ├── non_thermal
    │   ├── bremsstrahlung.py
    │   └── other.py
    └── thermal
        ├── continuum.py
        ├── lines.py
        └── other.py
```

binchensun:
Emm, I personally think the first "child" may be better to be categorized in terms of emission mechanisms: bremsstrahlung (free-free), line emission (free-bound, bound-bound), and others (e.g., pion decay). The difference between thermal and nonthermal is only related to the source electron/ion distribution. For example, bremsstrahlung can also take a thermal distribution, and a Kappa distribution bridges both thermal and nonthermal.

DanRyanIrish:
@binchensun Are you suggesting that we have one function/model that can produce both thermal/non-thermal bremsstrahlung spectra?  If so, what would the API look like?

binchensun:
> @binchensun Are you suggesting that we have one function/model that can produce both thermal/non-thermal bremsstrahlung spectra? If so, what would the API look like?
I have not yet thought about the detailed implementation. But a non-thermal thin-target bremsstrahlung takes in, e.g., a power-law, which normalizes upon n_i*V*n_e_nonthermal (if nonthermal electron distribution is described in density, not in flux), and thermal bremsstrahlung takes in T and EM (the latter has the same unit as n_i*V*n_e_nonthermal). The outputs are both in photons/(cm2 s keV). I guess it should be straightforward to put both the thermal and nonthermal thin-target modules into, e.g., bremsstrahlung.py? Looking at the SSWIDL implementation of thermal bremsstrahlung, it should be pretty straightforward: https://hesperia.gsfc.nasa.gov/ssw/packages/xray/idl/brem_49.pro

DanRyanIrish:
So here's an idea for the structure based on the thread here.  (It needs further discussion though.)  Modules/models would be based on emission mechanisms, rather than thermal/non-thermal.  (Although we could also have thermal/non-thermal models that are just wrappers around the relevant mechanisms.)
```
└── emission
    ├── bremsstrahlung.py
    ├── line.py
    ├── free-bound.py
    ├── other.py
    ├── thermal.py  # Wraps around relevant mechanisms and takes thermal distributions as inputs.
    ├── nonthermal.py  # Wraps around relevant mechanisms and takes nonthermal distributions as inputs.
└── particles
    ├── electrons  # Electron distributions
[...]

PR #151, opened 2024-06-13
Restructure repo.

DanRyanIrish:
## PR Description

This PR restructures the repo in line with the planned refactor.  It moves the `fitting_legacy` subpackage and all model-related modules to a new `legacy` subpackage and creates a new `models` subpackage.  Currently, this holds a few empty subpackages and modules to show how it should/could be structured going forward.

I suggest that we use this structure for our refactor, starting with the following:
* Create a `MatrixModel`
* Convert thermal function(s) to models
* Write a tutorial example using fake count and the scipy minimizer to fit a thermal spectrum
* Create `Spectrum` data container based on `NDCube` in a new `spectrum` subpackage, or decide that raw `NDCube`s are sufficient.
* Update tutorial to use the data container.
* Develop infrastructure required to simultaneously fit two count spectra, and write another tutorial showing how to do that.

There would be more to do, but this would actually prove the fundamental pillars of our refactor.

Other crucial work involves refactoring the non-thermal functions and making them models.


sunkit_spex/models/photon/thermal/tests/test_thermal.py
```diff
@@ -0,0 +1,406 @@
+import warnings
+
+import numpy as np
+import pytest
+
+import astropy.units as u
+
+from sunkit_spex import thermal
```
KriSun95, line 8: Does this need to be `from sunkit_spex.legacy import thermal`?

KriSun95: I'm happy to approve otherwise!

DanRyanIrish:
ping @samaloney

DanRyanIrish:
@settwi In merging your RHESSI PR, it seems to have caused an issue with my restructure PR.  Could you advise on how to handle the fact that `fitting_legacy.io` no longer includes `_read_rhessi_spec_file`.  Should it still be there?

sunkit_spex/models/photon/thermal/tests/test_thermal.py
DanRyanIrish: No.  It uses the imports from the top-level `__init__.py` file.  However, I realised that how that was done there was out of date and fixed it.  So I'll now mark this comment as resolved.

DanRyanIrish:
@KriSun95 If you're happy with this, can you officially approve it?

DanRyanIrish:
@samaloney It would be good to get your approval here.  The main things you should know about the new structure are:
* All old code is moved to `legacy`, including the old model modules and the `fitting_legacy` subpackage.
* A new `models` subpackage.  (Can you confirm you are happy with the name?)  It has the following structure:
  * `matrix.py` to hold Kris's future matrix model
  * `response.py` Models specific to responses.  I differentiate this from matrix.py because one could imagine a generla matrix model that isn't specific to a response.  If `response` turns out to be not needed because a generic matrix model is sufficient, this can be deleted in future.
  * `math.py` a module for mathematical models, e.g. a power-law.  (Again, is this the right module name?
  * `photon` subpackage for photon/physical models. Is this the right name?  Inside this so far is:
    * `thermal` subpackage for all code needed for thermal models.  Currently, it includes the `io.py` and `thermal.py` modules.  There are copies of these in the `legacy` subpackage so they can be refactored into proper models without losing the old functionality for now.
    * An empty `nonthermal` subpackage for future refactored nonthermal functions.

Mainly the names and the structure is what I'd like your views on before merging.

samaloney:
Over all I think this sounds good. I was just looking at some other existing packages and how they layout their models and some discussion we had before in #42.

* astropy: https://github.com/astropy/astropy/tree/main/astropy/modeling
* gammapy: https://github.com/gammapy/gammapy/tree/main/gammapy/modeling/models
* lmfit: https://github.com/lmfit/lmfit-py/blob/master/lmfit/models.py

**Questions:**
1. `models` or `modeling` don't really mind
2. Why move the io inside the model?

I think the integration stuff could live in it own module as it is essentially generic numerical integration.

My suggestion would be the following as I do like the idea of separating physically meaningful models from other mathematical or functionals ones.
```
.
└── models
    ├── models.py
    ├── instrument_response.py # don't need this now as Matrix would do
    └── physical.py  #these could be a directories/submodules if needed as you have 

```

but the imports could be like this (updating the case for class v function) 

```
from sunkit_spex.models.physical import thermal, thin_target, thick_target, warm_target
from sunlit_spex.models import Matrix
from sunkit_spex.models import Powerlaw
```

or if we find we have a lot of powerlaw or non_thermal models in the future we can change to

```
from sunkit_spex.models.powerlaw import MyFavPowerlaw
from sunkit_spex.models.pyhsical.nonthermal import WarmTarget, MediumHotTarget 😆 
```

Also moving stuff around the namespace while annoying is possible so we aren't going to be stuck if we find we need to change we can so it really doesn't matter too much so going to approve this PR.

DanRyanIrish:
>     1. `models` or `modeling` don't really mind

OK.  Let's stick with `models` then, since there's no clear convention.

>     2. Why move the io inside the model?

The thermal `io.py` was specifically for reading in CHIANTI tables for the thermal model.  There's another `io` in `fitting_legacy`.  There could be an `io` module for reading spectral files as well.  I agree this is messy.  But perhaps changing the name from `io` or moving it out later is an option too.

> ```
> .
> └── models
>     ├── models.py
>     ├── instrument_response.py # don't need this now as Matrix would do
>     └── physical.py  #these could be a directories/submodules if needed as you have 
> ```

OK.  I'll change `photon` to `physical`.  What would go in `models.py`?  Would this be equivalent to `mathematical.py`, i.e. `PowerLaw`?

> or if we find we have a lot of powerlaw or non_thermal models in the future we can change to
> 
> ```
> from sunkit_spex.models.powerlaw import MyFavPowerlaw
> from sunkit_spex.models.pyhsical.nonthermal import WarmTarget, MediumHotTarget 😆 
> ```

Even if there's a `sunkit_spex.models.physical.nonthermal` subpackage/module, we can use the `physical.__init__` to enable importing directly `from sunkit_spex.models.physical`.

DanRyanIrish:
Ah.  Do you mean that `models.py` would hold things like `PowerLaw` and `MatrixModel`?

samaloney:
> Ah. Do you mean that `models.py` would hold things like `PowerLaw` and `MatrixModel`? As in generic, non-physical, non-instrumental models?

Yea exactly anything that's not a physical model - if we find we have a too many we could create additional submodules/namespaces `response`, or `powerlaw` down the line.

DanRyanIrish:
OK, great.  I've made those changes.

samaloney (approved):
Once the notebooks run merge!

settwi:
why is the thermal code duplicated in legacy and models? @DanRyanIrish

settwi:
from Element:

why is the stix and rhessi “spec” code back in legacy? it’s kind of defunct now that we moved rhessi to “extern”

PR #152, opened 2024-06-15
Ruff lint and format, isort, codespell

samaloney:
Add ruff lint and format, sort and code spell to pre-commit and make all pass


settwi:
sweet!
if you run pre-commit locally does it now call ruff?

samaloney:
> if you run pre-commit locally does it now call ruff?

Yea that's the idea, so once this is merged it should.

samaloney:
The test fail is due to numpy2.0 removal so I think can be ignored for now

samaloney:
Again windows test fail is unrelated - maybe getting blocked by hesperia for too many downloads

ruff.toml
```diff
@@ -0,0 +1,42 @@
...
+  "build",
+  "sunkit_spex/version.py",
+  "sunkit_spex/legacy/fitting_legacy"
+]
+
+[lint]
+select = ["E", "F", "W", "UP", "PT"]
+extend-ignore = [
+  # pycodestyle (E, W)
+  "E501", # LineTooLong                             # TODO! fix
```
DanRyanIrish, line 15: What's this?

Same for lines below marked `# TODO! fix`

samaloney: Eh ignore line too long, pytext fixture as tuples and purest raise too broad errors until they are fixed in the code base as they can't be automatically fixed

KriSun95 (approved):
Wow! Clearly there are quite a few words I just can't spell when I'm typing quickly! This all seems good to me.

samaloney:
Ha I think if there was a typo award I would get it - worst thing this doesn't even catch them all.

PR #153, opened 2024-06-15
WIP Convert notebooks to gallery examples

samaloney:
Convert notebooks to gallery file so they can be run as part of the docs build.

There's a fair amount of overlap in the different examples should discuss what we should do, we could split into feature examples and papers recreations?

Also as far as I can see the current code doesn't always match what was in the tables 😬

* fitting_NuSTAR_spectra.py 🔴 
   * It runs and makes the plots but needs to be update to match rst, sphix-galery formatting
   * Very long
   * Some of the code causes error which is ok in notebook but will cause issues look at the `try ... except` blocks I had to add
* fitting_RHESSI_spectra.py 🟡  
   * Prob could do with some improved prose but formatting is good.
* fitting_attenuated_RHESSI_spectra.py 🟠
    * It runs and makes the plots but needs to be updated to match rst, sphix-galery formatting
* fitting_custom_spectra.py 🟠
   * It runs and makes the plots but needs to be updated to match rst, sphix-galery formatting


settwi:
is there a central spot we can put data files? i have some on a GDrive but that isn't exactly ideal

samaloney:
> is there a central spot we can put data files? i have some on a GDrive but that isn't exactly ideal

Eh good question, at the moment I don't think we have one. I just put them somewhere the individual files could be downloaded rather than a zip.

sunkit_spex/extern/rhessi.py
```diff
@@ -203,15 +205,18 @@ def _update_srm_state(self):
...
-                break
-
+        n_states = len(self._attenuator_state_info['states'])
+        new_att_state = self._attenuator_state_info['states'][0]  # default to first
+        if n_states > 1:
+            for i in range(n_states-1):
+                state = self._attenuator_state_info['states'][i]
+                if change_times[i] < start_time and end_time < change_times[i+1]:
+                    new_att_state = state
+                    break
```
samaloney, line 215: So when trying to use the older RHESSI example this broke this now work with both but not sure how robust it is.

sunkit_spex/extern/rhessi.py
```diff
@@ -691,15 +696,12 @@ def load_spectrum(spec_fn: str):
...
 def _extract_attenunator_info(att_dat) -> dict[str, list]:
     '''Pull out attenuator states and times'''
-    # swap the UNIX year with the date_obs year
-    # they don't match because the UNIX year is not the reference year
-    # as per Kim Tolbert
-    obs_year = atime.Time(att_dat.header['DATE_OBS']).datetime.year
-    dtimes = atime.Time(att_dat.data['SP_ATTEN_STATE$$TIME'], format='unix').datetime[0]
-    dtimes = [dt.replace(year=obs_year) for dt in dtimes]
+    n_attenuator_changes = att_dat.data['SP_ATTEN_STATE$$TIME'].size
+    atten_change_times = atime.Time(att_dat.data['SP_ATTEN_STATE$$TIME'], format='utime').utc
```
samaloney, line 700: So sunpy implemented the `ut` format so once its imported cause use directly

sunkit_spex/extern/rhessi.py
```diff
@@ -93,7 +95,7 @@ def load_prepare_spectrum_srm(self, spectrum_fn, srm_fn):
             self.start_data_time, self.end_data_time
 
         # Assume unattenuated SRM until spectroscopy interval is specified
-        UNATTENUATED_STATE = 0
+        UNATTENUATED_STATE = 1
```
samaloney, line 98: should prob default to first state?

sunkit_spex/extern/rhessi.py
```diff
@@ -1,6 +1,8 @@
 import copy
 import warnings
 
+import sunpy.time
```
samaloney, line 4: register `utime` with astropy Time

settwi:
the first state is state 0, 1 means thin attenuator insertedOn Jun 16, 2024, at 16:42, Shane Maloney ***@***.***> wrote:﻿
@samaloney commented on this pull request.



In sunkit_spex/extern/rhessi.py:
> @@ -93,7 +95,7 @@ def load_prepare_spectrum_srm(self, spectrum_fn, srm_fn):
             self.start_data_time, self.end_data_time
 
         # Assume unattenuated SRM until spectroscopy interval is specified
-        UNATTENUATED_STATE = 0
+        UNATTENUATED_STATE = 1

should prob default to first state?

sunkit_spex/extern/rhessi.py
settwi: the first state is state 0, 1 means thin attenuator inserted

sunkit_spex/extern/rhessi.py
settwi: what is utime?

samaloney: UT seconds from 1979-01-01 00:00:00 UTC, ignoring leap seconds (https://docs.sunpy.org/en/stable/generated/api/sunpy.time.TimeUTime.html)

sunkit_spex/extern/rhessi.py
samaloney: Sure but in the other RHESSI example SRM its 1 so needs to work with either.

DanRyanIrish (approved):
Approve pending resolving conflicts and getting tests to pass.

PR #161, opened 2024-09-12
Add albedo correction

### LINKED ISSUE #128: Add albedo correction
### Describe the feature

A few people have been asking if we have implemented the albedo correction similar to OSPEX

### Proposed solution

_No response_

samaloney:
## PR Description

A a function to read in precomputed greens table and return spectrum with albedo correction, would close #128

- [x] Split function at the moment every time the function is called we would make web request and load a file from disk both very slow not ideal for fitting.   The idea would be split up into at least two parts depending on the use case/s. As far as I know often $\theta$ is fix so the loading and interpolation between $\theta$ can be cached. So one function to load the appropriate greens matrix file and interpolate for given angle. A second function would call the previous function and do the energy intepotalation (as this could change in a fit e.g. gain correction) and fold the given spectrum through obtained green matrix.
- [x] Turn in to astropy model e.g. `spec | albedo` (still waiting on proper units support in astroy for this work https://github.com/astropy/astropy/issues/17302)
- [x] Docs and tests

Recreated Fig 2 upper from the paper 

albedo_fig

### DIFF
--- a/sunkit_spex/models/physical/albedo.py
+++ b/sunkit_spex/models/physical/albedo.py
@@ -80,7 +80,9 @@ class Albedo(FittableModel):
         description="Angle between the observer and the source",
         fixed=False,
     )
-    anisotropy = Parameter(name="anisotropy", default=1, description="The anisotropy used for albedo correction", fixed=True)
+    anisotropy = Parameter(
+        name="anisotropy", default=1, description="The anisotropy used for albedo correction", fixed=True
+    )
 
     _input_units_allow_dimensionless = True
 
--- a/sunkit_spex/models/physical/albedo.py
+++ b/sunkit_spex/models/physical/albedo.py
@@ -13,6 +13,11 @@ def albedo(spec, energy: Quantity[u.keV], theta: Quantity[u.deg, 0, 90], anisotr
     r"""
     Add albedo correction to input spectrum
 
+    Correct input model spectrum for the component reflected by the solar atmosphere following [Kontar20006]_ using
+    precomputed green matrices from SSW.
+
+    .. [Kontar20006] https://doi.org/10.1051/0004-6361:20053672
+
     Parameters
     ----------
     spec :
@@ -32,12 +37,12 @@ def albedo(spec, energy: Quantity[u.keV], theta: Quantity[u.deg, 0, 90], anisotr
     >>> e = np.linspace(5,  500, 1000)
     >>> e_c = e[:-1] + np.diff(e)
     >>> s = 125*e_c**-3
-    >>> corrected = albedo(s, energy, theta=45*u.deg)
+    >>> corrected = albedo(s, e, theta=45*u.deg)
     """
     base_url = "https://soho.nascom.nasa.gov/solarsoft/packages/xray/dbase/albedo/"
     mu = np.cos(theta)
 
-    # what bout 0 and 1 i assume so close to 05 and 95 that it doesn't matter
+    # what about 0 and 1 assume so close to 05 and 95 that it doesn't matter
     # load precomputed green matrices
     if 0.5 <= mu <= 0.95:
         low = 5 * np.floor(mu * 20)
@@ -71,9 +76,9 @@ def albedo(spec, energy: Quantity[u.keV], theta: Quantity[u.deg, 0, 90], anisotr
     interp = RegularGridInterpolator((energy_grid_centers.to_value(u.keV), energy_grid_centers.to_value(u.keV)), albedo)
 
     de = np.diff(energy)
-    engery_centers = energy[:-1] + de / 2
+    energy_centers = energy[:-1] + de / 2
 
-    X, Y = np.meshgrid(engery_centers.to_value(u.keV), engery_centers.to_value(u.keV))
+    X, Y = np.meshgrid(energy_centers.to_value(u.keV), energy_centers.to_value(u.keV))
     albedo_interp = interp((X, Y))
 
     albedo_interp = albedo_interp * de.value / anisotropy
--- a/sunkit_spex/models/physical/albedo.py
+++ b/sunkit_spex/models/physical/albedo.py
@@ -11,7 +11,7 @@ from astropy.units import Quantity
 
 from sunpy.data import cache
 
-__all__ = ["Albedo", "get_albedo_matrix"]
+__all__ = ["Albedo"]
 
 
 class Albedo(FittableModel):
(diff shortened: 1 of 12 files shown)


edkontar:
I have found IDL code used in 2005 to generate Green Matrices for OSPEX and put it on github:

https://github.com/edkontar/albedo

natsat0919:
I have added the changes we made to the albedo model during the workshop. For now we made the model "unitless" until we resolve the astropy model unit problems with SRM. All the units-related lines are commented so it can be easily updated.

sunkit_spex/models/physical/albedo.py
```diff
@@ -87,7 +87,7 @@ def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
 
     def evaluate(self, spectrum, theta, anisotropy):
-        albedo_matrix = get_albedo_matrix(self.energy_edges, theta, anisotropy)
+        albedo_matrix = get_albedo_matrix(self.energy_edges, self.theta, anisotropy)
```
samaloney, line 90: This may work but I think it breaks the API convention for `evaluate` as `evaluate` should not use any parameters on the instance only this passed as argument

jajmitchell: Ahh, this was added in to allow fitting with Scipy, but as you rightly point out this is probably not the correct way forward, for the astropy fitting to work we'd need to take a different approach anyway

edkontar:
In OSPEX/drm_correct_albedo.pro, the anisotropy parameter is “a coefficient showing the ratio of the flux in observer direction to the flux downwards." A smaller value means more of an effect from albedo.  A value of 1000 means very little effect. if anisotropy=1 (default) the source is isotropic. However, in the paper, the coefficient α that accounts for anisotropy of the source in the Eddington approximation – that is, "distinct but constant specific intensities in the downward and in the upward hemispheres, in the ratio α” i.e. the anisotropy described as the reciprocal of the anisotropy parameter in the OSPEX.

Should the description state the definition of "α" was assumed as in OSPEX?

class Albedo(FittableModel):
    r"""
    Aldedo model which adds albdeo correction to input spectrum.
    Following [Kontar2006]_ using precomputed green matrices distributed as part of [SSW]_.
    .. [Kontar2006] https://doi.org/10.1051/0004-6361:20053672
    .. [SSW] https://www.lmsal.com/solarsoft/
    Parameters
    ==========
    energy_edges :
        Energy edges associated with input spectrum
    theta :
        Angle between Sun-observer line and X-ray source
    anisotropy :
        Ratio of the flux in observer direction to the flux downwards, 1 for an isotropic source

sunkit_spex/models/physical/albedo.py
```diff
@@ -0,0 +1,190 @@
...
+
+    _input_units_allow_dimensionless = True
+
+    def __init__(self, theta=u.Quantity(theta.default, theta.unit),
+                 anisotropy=anisotropy.default,
+                 **kwargs):
+
+        self.energy_edges = kwargs.pop("energy_edges")
+
+        self._get_green_matrix(theta)
```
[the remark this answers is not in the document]
jajmitchell: Perhaps we can remove the decorator then? referencing in the __init__ step seems to be more inline with the Astropy API. From looking over the Astropy API all model classes have their reliant functions handled internally, the previous method was encountering issues with fitting theta as a free parameter, I think at least in part because of the functions being defined outside of the class, although I could be wrong about that. Either way, this version solves that issue and fully functions with the Astropy fitters, and seemingly produces correct results, also it is more in line with the Astropy model class API. I guess it comes back to the main question of to what degree we want to align ourselves with the API

jajmitchell:
I have reverted the code back to its state before the previous commit with some small edits, the functions are now defined outside of the class and the class is used a wrapper, therefore the tests should now work, is it ok for me to push these changes?

samaloney:
Go for it, sounds good

sunkit_spex/models/physical/albedo.py
```diff
@@ -0,0 +1,236 @@
...
+    albedo_interp = albedo_interpolator((X, Y))
+
+    # Scale by anisotropy
+    albedo_interp = (albedo_interp * de) / anisotropy
+
+    # Take a transpose
+    return albedo_interp.T
+
+
+def get_albedo_matrix(energy_edges: Quantity[u.keV], theta: Quantity[u.deg], anisotropy=1):
```
samaloney, line 200:
```suggestion
@quantity_input
def get_albedo_matrix(energy_edges: Quantity[u.keV], theta: Quantity[u.deg], anisotropy=1):
```

sunkit_spex/models/physical/albedo.py
```diff
@@ -0,0 +1,236 @@
...
+                         **kwargs)
+
+    
+
+    def evaluate(self, spectrum, theta, anisotropy):
+
+        if hasattr(theta, "unit"):
+            theta = Quantity(theta.value,theta.unit)
+        else:
+            theta = theta*u.deg
```
samaloney, line 102:
```suggestion
        if not isinstance(theta, Quantity)
            theta = theta*u.deg
```

settwi: hey it looks like y'all are making good progress on this, i am gonna put a couple style suggestions along these lines if that's ok. feel free to reject them if you want but yea

sunkit_spex/models/physical/albedo.py
```diff
@@ -0,0 +1,235 @@
...
+        self.energy_edges = kwargs.pop("energy_edges")
+
+        super().__init__(*args,
+                         **kwargs)
+
+    
+
+    def evaluate(self, spectrum, theta, anisotropy):
+
+        if not isinstance(theta, Quantity)
```
samaloney, line 99:
```suggestion
        if not isinstance(theta, Quantity):
```

sunkit_spex/models/physical/albedo.py
```diff
@@ -0,0 +1,235 @@
...
+    albedo_interp = albedo_interpolator((X, Y))
+
+    # Scale by anisotropy
+    albedo_interp = (albedo_interp * de) / anisotropy
+
+    # Take a transpose
+    return albedo_interp.T
+
+
+@quantity_input
```
samaloney, line 198:
```suggestion
@u.quantity_input
```

jajmitchell: need to change to @u.quantity_input

samaloney:
Ok so I think this is good to go,  @natsat0919 and @jajmitchell could you review?

jajmitchell (approved):
I think that this branch is ready to be merged. I am still trying to get the albedo model to work with a matrix model in astropy class format and with the thermal model in astropy format, which may require some fiddling with the unit handling, we could delay the merge here and get the unit handling all working together before merging, or merge and then deal with this in a differently titled pull request as the albedo model does function now as a stand alone and this is a wider problem. My instinct would be to merge, then open a "compound model unit handling" pull request or issue, but am easy either way.

jajmitchell:
@samaloney, just realised that the anisotropy parameter needs to have the "name" argument added in its definition, apologies that I spotted this after adding my review, are you happy for me to push this minor change then re-review?

sunkit_spex/models/physical/albedo.py
```diff
@@ -0,0 +1,228 @@
...
+    theta = Parameter(
+        name="theta",
+        default=0,
+        unit=u.deg,
+        min=-90,
+        max=90,
+        description="Angle between the observer and the source",
+        fixed=False,
+    )
+    anisotropy = Parameter(default=1, description="The anisotropy used for albedo correction", fixed=True)
```
samaloney, line 83:
```suggestion
    anisotropy = Parameter(name="anisotropy", default=1, description="The anisotropy used for albedo correction", fixed=True)
```

samaloney:
pre-commit.ci autofix

natsat0919 (approved):
I played around with the code and all seems to work well

PR #181, opened 2025-01-27
make RhessiLoader compatible with legacy fitter args

### LINKED ISSUE #180: RHESSI loader has incompatibilities with "legacy" portion of fitting code
### Describe the bug

When the [`RhessiLoader`](https://github.com/sunpy/sunkit-spex/blob/main/sunkit_spex/extern/rhessi.py) got moved to `extern` some of the methods/properties got renamed or modified.

Thanks to @ianan for pointing this out. The "legacy" code should continue to work even if the loader has been moved to `extern`

Here's a list of things that we need to support. Most can be fixed with simple patches. The old names should probably be given deprecation warnings.
- [x] Passing in the spectrum and SRM FITS files to `Fitter` doesn't work for RHESSI
- [x] the property `data2data_minus_background` got renamed to `background_subtract`
- [x] Time-related properties
    - [x] `start_event_time`
    - [x] `end_event_time`
    - [x] `start_background_time`
    - [x] `end_background_time`
    - Frankly--I think these should be updated in pairs, but we can support the individual getters/setters
- [x] Spectrum energy bounds don't get respected when set near the edges for plotting
    - Not sure if this was ever the case, but we can support it here if desired.

I think those are the changes between the previous RHESSI loader and the current. I'll open a PR shortly which addresses these.

settwi:
Fixes #180

### DIFF
--- a/changelog/181.bugfix.rst
+++ b/changelog/181.bugfix.rst
@@ -1,0 +1,1 @@
+Add back legacy parameters/functionality relating to `~sunkit_spex.extern.stix.RhessiLoader`.
--- a/examples/fitting_RHESSI_spectra.py
+++ b/examples/fitting_RHESSI_spectra.py
@@ -118,20 +118,6 @@ plt.rcParams["font.size"] = default_text
 
 #####################################################
 #
-# An alternative to set the event and background times like above is by using the `select_time` method. E.g.,
-#
-# .. code-block::python
-#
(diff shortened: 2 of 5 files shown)


settwi:
@ianan can you try running your script with this branch? you can install it like so:
```bash
pip install git+https://github.com/settwi/sunkit-spex.git@rhessi-legacy-compatibility
```

samaloney:
I'd say this does need change log?

settwi:
> I'd say this does need change log?

fair enough :p

samaloney (approved):

KriSun95 (approved):

PR #184, opened 2025-01-30
Defer thermal model setup until actually needed

samaloney:
This code change the way the thermal module configuration is loaded it uses a class to defer the process until the props are actually needed and also caches them.

At the moment I haven't implemented a setter as I don't know how to verify the data would be in the correct format.

### DIFF
--- a/sunkit_spex/models/physical/thermal.py
+++ b/sunkit_spex/models/physical/thermal.py
@@ -191,9 +191,9 @@ def setup_default_abundances(filename=None):
 
 
 # Read line, continuum and abundance data into global variables.
-CONTINUUM_GRID = setup_continuum_parameters()
-LINE_GRID = setup_line_parameters()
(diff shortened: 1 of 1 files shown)


jajmitchell:
@samaloney I have also been working on the thermal model on the branch https://github.com/jajmitchell/sunkit-spex/tree/thermal-model-astropy-wrapper, still a bit messy, and definitely needs streamlining, but the wrapper seems to work with unit handling and produces the same results as the thermal_emission model. I was planning on opening a PR on this, but there would be lots of conflict with this one, how should we go forward with it?

samaloney:
@jajmitchell  I'd prob just open the PR anyway, conflicts can't really be helped sometimes

samaloney:
I think this could actually be better accomplished with [custom module attributes](https://docs.python.org/3/reference/datamodel.html#customizing-module-attribute-access)

PR #189, opened 2025-02-14
Adds Astropy wrapper to the thermal model 

jajmitchell:
Adds an astropy model class wrapper to the thermal model

This PR changes the format of the thermal model by adding an astropy class wrapper, ThermalEmission. 

Currently the model can be evaluated both alone and when combined with the albedo model. Below is a plot showing the results of evaluating the model both alone and with the albedo model:

Thermal_albedo

The model can be fit with the standard astropy fitting packages as a standalone. 

The following needs to be changed before merging:

- [ ] sanitize inputs and range checks need to be reimplemented, and in such a fashion that they don't interfere with the unit stripping inherent to astropy fitting.
- [ ] Tests need to be written.

### DIFF
--- a/sunkit_spex/models/physical/thermal.py
+++ b/sunkit_spex/models/physical/thermal.py
@@ -1,3 +1,5 @@
+from functools import lru_cache
+
 import copy
 import warnings
 
@@ -5,6 +7,12 @@ import numpy as np
 from scipy import interpolate, stats
 
 import astropy.units as u
+
+import astropy.units as u
+from astropy.modeling import FittableModel, Parameter
+from astropy.units import Quantity
+
+from sunpy.data import cache
 
 from sunpy.data import manager
 
@@ -15,12 +23,13 @@ from sunkit_spex.models.physical.io import (
 )
 
 __all__ = [
+    "ThermalEmission",
+    "thermal_emission",
     "continuum_emission",
     "line_emission",
     "setup_continuum_parameters",
-    "setup_default_abundances",
     "setup_line_parameters",
-    "thermal_emission",
+    "setup_default_abundances",
 ]
 
 doc_string_params = """
@@ -71,6 +80,199 @@ Returns
 flux: `astropy.units.Quantity`
     The photon flux as a function of temperature and energy.
 """
+
+class ThermalEmission(FittableModel):
+    r"""
+    """
+    n_inputs = 1
+    n_outputs = 1
+    
+    temperature = Parameter(
+        name="temperature",
+        default=1e7,
+        min=1e6,
+        max=1e8,
+        unit=u.K,
+        description="Temperature of the plasma",
+        fixed=False,
+    )
+    
+    emission_measure = Parameter(
+        name="emission_measure",
+        default=1e8, 
+        unit=(u.cm ** (-3)),
+        description="Emission measure of the observer", 
+        fixed=False)
+    
+    observer_distance = Parameter(
+        name="observer_distance",
+        default=1,
+        min = 0.01,
+        unit=u.AU,
+        description="Distance to the observer",
+        fixed=True) 
+
+    input_units_equivalencies = {'keV': u.spectral(),'K':u.temperature_energy()}
+    _input_units_allow_dimensionless = True
+
+    def __init__(self,
+                  temperature=u.Quantity(temperature.default,temperature.unit), 
+                  emission_measure=u.Quantity(emission_measure.default,emission_measure.unit), 
+                  observer_distance=u.Quantity(observer_distance.default,observer_distance.unit),
+                  abundance_type="sun_coronal_ext",
+                  relative_abundances=None,
+                 **kwargs):
+
+        self.energy_edges = kwargs.pop("energy_edges")
+        self.abundance_type = abundance_type
+        self.relative_abundances = relative_abundances
+
+        super().__init__(temperature=temperature,emission_measure=emission_measure,observer_distance=observer_distance,
+                         **kwargs)
+
+    def evaluate(self, spectrum, temperature, emission_measure, observer_distance):
+
+        # if not isinstance(temperature, Quantity):
+        #     temperature = temperature * u.K
+
+        # if not isinstance(emission_measure, Quantity):
+        #     emission_measure = emission_measure * (u.cm ** (-3))
+
+        # if not isinstance(observer_distance, Quantity):
+        #     observer_distance = (observer_distance * u.AU).to(u.cm)
+        # else:
+        #     observer_distance = observer_distance.to(u.cm)
+
+        flux = thermal_emission(self.energy_edges,
+            temperature,
+            emission_measure,
+            observer_distance,
+            self.abundance_type,
+            self.relative_abundances
+            )
+        
+        if hasattr(temperature, "unit"):
+            return flux
+        else:
+            return flux.value
+
+    @property
+    def input_units(self):
+        # The units for the 'energy_edges' variable should be an energy (default keV)
+        return {self.inputs[0]: u.keV}
+
+    @property
+    def return_units(self):
+        return {self.outputs[0]: u.ph / u.keV * u.s**-1 *u.cm**-2}
+
+
+    def _parameter_units_for_data_units(self, inputs_unit, outputs_unit):
+        return {"temperature": u.K,"emission_measure":(u.cm ** (-3)),"observer_distance":u.AU}
+    
+
+# class ContinuumEmission(FittableModel):
+#     r"""
+#     """
+#     n_inputs = 1
+#     n_outputs = 1
+#     temperature = Parameter(
+#         name="temperature",
+#         default=0,
+#         unit=u.K,
+#         description="Temperature of the plasma",
+#         fixed=False,
+#     )
+    
+#     emission_measure = Parameter(
+#         name="emission_measure",
+#         default=1, 
+#         unit=(u.cm ** (-3), u.cm ** (-5)),
+#         description="Emission measure of the observer", 
+#         fixed=False)
+    
+#     observer_distance = Parameter(
+#         name="observer_distance",
+#         default=1,
+#         unit=u.cm,
+#         description="Distance to the observer"
+#     ) 
+
+#     _input_units_allow_dimensionless = True
+
+#     def __init__(self, *args,
+#                  **kwargs):
+
+#         self.energy_edges = kwargs.pop("energy_edges")
+
+#         super().__init__(*args,
+#                          **kwargs)
+
+#     def evaluate(self, spectrum, theta, anisotropy):
+
+#         if hasattr(theta, "unit"):
+#             theta = Quantity(theta.value,theta.unit)
+#         else:
+#             theta = theta*u.deg
+ 
+#         albedo_matrix = get_albedo_matrix(self.energy_edges, theta, anisotropy)
+
+#         return spectrum + spectrum @ albedo_matrix
+
+#     def _parameter_units_for_data_units(self, inputs_unit, outputs_unit):
+#         return {"theta": u.deg}
+
+
+# class LineEmission(FittableModel):
+#     r"""
+#     """
+#     n_inputs = 1
+#     n_outputs = 1
+#     temperature = Parameter(
+#         name="temperature",
+#         default=0,
+#         unit=u.K,
+#         description="Temperature of the plasma",
+#         fixed=False,
+#     )
+    
+#     emission_measure = Parameter(
+#         name="emission_measure",
+#         default=1, 
+#         unit=(u.cm ** (-3), u.cm ** (-5)),
+#         description="Emission measure of the observer", 
+#         fixed=False)
+    
+#     observer_distance = Parameter(
+#         name="observer_distance",
+#         default=1,
+#         unit=u.cm,
+#         description="Distance to the observer"
+#     ) 
+
+#     _input_units_allow_dimensionless = True
+
+#     def __init__(self, *args,
+#                  **kwargs):
+
+#         self.energy_edges = kwargs.pop("energy_edges")
+
+#         super().__init__(*args,
+#                          **kwargs)
+
+#     def evaluate(self, spectrum, theta, anisotropy):
+
+#         if hasattr(theta, "unit"):
+#             theta = Quantity(theta.value,theta.unit)
+#         else:
+#             theta = theta*u.deg
+ 
+#         albedo_matrix = get_albedo_matrix(self.energy_edges, theta, anisotropy)
+
+#         return spectrum + spectrum @ albedo_matrix
+
+#     def _parameter_units_for_data_units(self, inputs_unit, outputs_unit):
+#         return {"theta": u.deg}
+
 
 
 def setup_continuum_parameters(filename=None):
@@ -197,16 +399,17 @@ DEFAULT_ABUNDANCES = setup_default_abundances()
 DEFAULT_ABUNDANCE_TYPE = "sun_coronal_ext"
 
 
-@u.quantity_input(
-    energy_edges=u.keV, temperature=u.K, emission_measure=(u.cm ** (-3), u.cm ** (-5)), observer_distance=u.cm
-)
+# @u.quantity_input(
+#     energy_edges=u.keV, temperature=u.K, emission_measure=(u.cm ** (-3), u.cm ** (-5)), observer_distance=u.cm
+# )
+@u.quantity_input
 def thermal_emission(
     energy_edges,
     temperature,
     emission_measure,
+    observer_distance,
     abundance_type=DEFAULT_ABUNDANCE_TYPE,
-    relative_abundances=None,
-    observer_distance=(1 * u.AU).to(u.cm),
+    relative_abundances=None
 ):
     f"""Calculate the thermal X-ray spectrum (lines + continuum) from the solar atmosphere.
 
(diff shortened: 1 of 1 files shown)


samaloney:
Looks like it nearly ready, nice work. What's up with the bumps in the albedo model?

settwi:
the bumps look like they might be due to an interpolation issue---for things like [mass attenuation coefficients](https://physics.nist.gov/PhysRefData/XrayMassCoef/intro.html), they need to be interpolated in (log(x), log(y)) space to keep the resulting log-log plots smooth. is the albedo data interpolated in that way?

jajmitchell:
@natsat0919 and I realised that the green matrices have a resolution of 1keV, which is whats causing the strange patterns. This issue is present in the legacy code, however it wouldn't be apparent when using STIX data. To solve this, we should run off some higher resolution tables. @samaloney would you be able to host them?

sunkit_spex/models/physical/thermal.py
```diff
@@ -72,6 +81,199 @@
...
+
+    def __init__(self,
+                  temperature=u.Quantity(temperature.default,temperature.unit), 
+                  emission_measure=u.Quantity(emission_measure.default,emission_measure.unit), 
+                  observer_distance=u.Quantity(observer_distance.default,observer_distance.unit),
+                  abundance_type="sun_coronal_ext",
+                  relative_abundances=None,
+                 **kwargs):
+
+        self.energy_edges = kwargs.pop("energy_edges")
```
settwi, line 122: should `energy_edges` be an instance variable? i thought they got passed to `evaluate`

jajmitchell: good point, I have changed this now, will push the changes soon along with some others.

sunkit_spex/models/physical/thermal.py
```diff
@@ -95,18 +95,64 @@ class ThermalEmission(FittableModel):
...
         min = 0.01,
         unit=u.AU,
         description="Distance to the observer",
-        fixed=True) 
+        fixed=True
+    )
+
+    mg = Parameter(
+        name="Mg",
+        default=8.15,
```
natsat0919, line 115: I’m almost certain that these default values are not in a correct format. The relative abundances in the files that are loaded in sunkit-spex in are in form of something like: elementX_abundance = 10**(X_abundance ) / 10**(H_abundance). So the abundances are very small numbers. Maybe it’s worth to set them to values that are loaded in the `abundance_type` column. This is especially important if a user chooses different abundance table, doesn’t want to vary the Mg, Si, S… abundances and might not necessarily be aware that we hardcoded some default values that aren’t from the `abundance_type` column they selected. In this case the default values would be incorrect.

jajmitchell: I took the values from here https://hesperia.gsfc.nasa.gov/ssw/packages/xray/dbase/chianti/sun_coronal_abund.txt, which is the file that is loaded I believe, they are also consistent with the values with Xspec models that use the Chianti tables and the model still looks sensible when evaluated with these values. Where is it you see the smaller values? 

 I agree though that the defaults should be set to whatever the default is in the selected abundance file, i will try to add this in, perhaps not as an explicit default as it is now but set in the init function.

natsat0919: What do you get when you try to print the abundances? e.g.:
```
abundances = thermal.setup_default_abundances()
print(abundances['sun_coronal_ext'])
```

jajmitchell: Ahh, sorry, yes you're right! Hmm, it would be useful if we could reformat the outputs to match the values I have inserted as the defaults. I'll have a look and push when I have something that works

natsat0919: It is an easy conversion: `abundance_X = 12 + np.log10(abundance_X_from_the_table)`. Might be worth to change the model input abundances to the format of the ones in the table inside the `calculate_abundances` function.

jajmitchell: Think I've fixed the abundances, I have kept the defaults as they are, but they get updated if the user specifies a different table, also sorted the units issue. probably needs tidying up a little, but I think it works

sunkit_spex/models/physical/thermal.py
```diff
@@ -170,6 +170,17 @@ def __init__(self,
 
         self.abundance_type = abundance_type
 
+        abundances = setup_default_abundances()
+
+        abundances = abundances[abundance_type]
+
+        if abundance_type != "sun_coronal_ext":
+            mg = 12+np.log10(abundances[11])
```
natsat0919, line 178: We should also include aluminium in the abundance list as it is commonly used as a free parameter as well: e.g. https://arxiv.org/pdf/2107.07825

sunkit_spex/models/physical/thermal.py
```diff
@@ -1078,38 +1082,18 @@ def _error_if_low_energy_input_outside_valid_range(input_values, grid_range, par
...
     abundances = DEFAULT_ABUNDANCES[abundance_type].data
 
-    abundances[12] = mg
-    abundances[14] = si
-    abundances[16] = s
-    abundances[18] = ar
-    abundances[20] = ca
-    abundances[26] = fe
-
-    # if relative_abundances:
```
natsat0919: Might be worth to keep this code for relative abundances? Not sure what others think

sunkit_spex/models/physical/thermal.py
jajmitchell: sorted it :)

jajmitchell:
I want to start on the thick target model also, should I do that in this PR, or open a separate one using this same branch? It is important that I have the ThermalEmission class in the same branch as I want to test the model combination.

natsat0919:
> I want to start on the thick target model also, should I do that in this PR, or open a separate one using this same branch? It is important that I have the ThermalEmission class in the same branch as I want to test the model combination.

I would suggest to start a new branch. Both models are quite a lot of work with a lot of code so it would be best to keep them separate

samaloney:
> I want to start on the thick target model also, should I do that in this PR, or open a separate one using this same branch? It is important that I have the ThermalEmission class in the same branch as I want to test the model combination.

I'd suggest a separate branch/PR - each PR should ideally be an independent/separate feature.

sunkit_spex/models/physical/thermal.py
```diff
@@ -752,27 +1313,28 @@ def _warn_if_input_outside_valid_range(input_values, grid_range, param_name, par
...
+def _error_if_low_energy_input_outside_valid_range(input_values, grid_range, param_name, param_unit):
+    if input_values.min() < grid_range[0]:
+        message = (
+            f"Lower bound of the input {param_name} must be within the range "
+            f"{grid_range[0]}--{grid_range[1]} {param_unit}. "
+        )
+        raise ValueError(message)
+
+
+def _calculate_abundances(abundance_type, mg, al, si, s,
```
KriSun95, line 1325: There are a few changes here. Would it be a good idea to maintain the checks that ensure the abundances are valid and positive?

sunkit_spex/models/physical/thermal.py
```diff
@@ -72,6 +79,466 @@
     The photon flux as a function of temperature and energy.
 """
 
+class ThermalEmission(FittableModel):
```
KriSun95, line 82: Would it make more sense to create this model using the `LineEmission` and `ContinuumEmission` models below? Open to suggestions here, but I think this would avoid duplicating code and make it easier to maintain/debug (e.g., if abundances for more elements were to be included).

sunkit_spex/models/physical/thermal.py
```diff
@@ -323,11 +839,19 @@ def _continuum_emission(energy_edges_keV, temperature_K, abundances):
         The abundances for the all the elements.
     """
     # Handle inputs and derive some useful parameters from them
-    log10T_in = np.log10(temperature_K)
-    T_in_keV = temperature_K / 11604518  # Convert temperature from K to keV.
+    if isinstance(temperature_K, Quantity):
+        temperature_K = temperature_K
```
KriSun95, line 843: Remove this line?

sunkit_spex/models/physical/thermal.py
```diff
@@ -323,11 +839,19 @@ def _continuum_emission(energy_edges_keV, temperature_K, abundances):
...
+        log10T_in = np.log10(temperature_K.value)
+        T_in_keV = temperature_K.to("keV", equivalencies=u.temperature_energy())
+        energy_edges_keV = energy_edges_keV.to(u.keV)
+    else:
```
KriSun95, line 847: I'm wondering if this `it`/`else` should be removed. The doc-string says that `temperature_K` "must not be a scalar" so the `isinstance(temperature_K, Quantity)` should always be `True` for expected behaviour. Should we replace this with a check like:
```
if not isinstance(temperature_K, Quantity):
    raise TypeError("`temperature_K` should be an Astropy `Quantity` object, like `u.K` ya silly goose!")

log10T_in = np.log10(temperature_K.value)
T_in_keV = temperature_K.to("keV", equivalencies=u.temperature_energy())
energy_edges_keV = energy_edges_keV.to(u.keV)
```

sunkit_spex/models/physical/tests/test_thermal.py
```diff
@@ -592,34 +592,34 @@ def test_line_emission_against_ssw(ssw):
 
 def test_scalar_energy_input():
```
KriSun95, line 593: I think these tests might need more fixing with the addition of the abundances as inputs to the `thermal.thermal_emission` model. I think this is the cause of one of the failing checks for the PR.

sunkit_spex/models/physical/thermal.py
```diff
@@ -72,6 +79,466 @@
...
+                  emission_measure=u.Quantity(emission_measure.default,emission_measure.unit), 
+                  observer_distance=u.Quantity(observer_distance.default,observer_distance.unit),
+                  mg=mg.default,
+                  al=al.default,
+                  si=si.default,
+                  s=s.default,
+                  ar=ar.default,
+                  ca=ca.default,
+                  fe=fe.default,
+                  abundance_type="sun_coronal_ext",
```
KriSun95, line 178: For the doc-string perhaps, but we should note somewhere that the individual elemental abundances take priority over any input to the `abundance_type` input. At least, that's what I think is the priority when following the code.

sunkit_spex/models/physical/thermal.py
```diff
@@ -72,6 +79,466 @@
...
+        fixed=True
+    )
+
+    input_units_equivalencies = {'keV': u.spectral(),'K':u.temperature_energy()}
+    _input_units_allow_dimensionless = True
+
+    def __init__(self,
+                  temperature=u.Quantity(temperature.default,temperature.unit), 
+                  emission_measure=u.Quantity(emission_measure.default,emission_measure.unit), 
+                  observer_distance=u.Quantity(observer_distance.default,observer_distance.unit),
```
KriSun95, line 170: We chatted about this earlier about moving the distance scaling out to another model that can be applied independently which sounded good.

sunkit_spex/models/physical/thermal.py
```diff
@@ -719,15 +1262,33 @@ def _weight_emission_bins(
...
+
     if energy_edges.isscalar or len(energy_edges) < 2 or energy_edges.ndim > 1:
         raise ValueError("energy_edges must be a 1-D astropy Quantity with length greater than 1.")
-    energy_edges_keV = energy_edges.to_value(u.keV)
-    temperature_K = temperature.to_value(u.K)
+
+    if isinstance(energy_edges, Quantity):
+        energy_edges_keV = energy_edges.to(u.keV)
+    else:
+        energy_edges_keV = energy_edges
```
KriSun95, line 1273: I'm wondering if this is the behaviour we want. We want the values to be in keV regardless of the input quantity so should we not always want to cast it (e.g., `<< u.keV`). I'm mainly concerned with the line that leaves the input alone if the input isn't a `Quantity`.

KriSun95 (changes requested):
There are a few things I'm curious about here but this is looking great! 

For completeness, I think it would be good to add a "changelog" and also run `tox` or `pre-commit` locally on the branch to try and get the PR checks to pass.

sunkit_spex/models/physical/thermal.py
jajmitchell: Rather than using the existing checks, I think we should provide "reasonable" limits to each parameter, which in itself would ensure the values were positive whilst not interfering with the fitting process.

sunkit_spex/models/physical/thermal.py
jajmitchell: It is true that this way there is some duplication, however currently, the thermal, line and continuum can all be evaluated and fit separately using the astropy and scipy fitters, for this reason I'd opt to leave it how it is, but am open to arguments as to why we should streamline

sunkit_spex/models/physical/thermal.py
jajmitchell: If you specifically mean this line `temperature_K = temperature_K` , then agreed, it is a very silly line indeed

sunkit_spex/models/physical/thermal.py
jajmitchell: This `if/else` statement is needed, as when fitting the units are by default stripped, so this function needs to work with both scalar and quantity inputs, which is what this `if/else` statement facilitates

sunkit_spex/models/physical/tests/test_thermal.py
jajmitchell: Yeah, definitely need to do some work on the tests, including the inclusion of the IDL benchmarking discussed at the last meeting

sunkit_spex/models/physical/thermal.py
jajmitchell: So the way it works currently, is that the defaults are updated in the initialisation if a different abundance_type is specified

sunkit_spex/models/physical/thermal.py
jajmitchell: As with the `if/else` statement discussed above, it is essential for the fitting that these functions work with no units, perhaps we could throw an error in the initialise step if the energy_edges are not in kev, or a compatible unit?

sunkit_spex/models/physical/thermal.py
jajmitchell: I have now removed the observer_distance from this model, and created a separate model class that can scale any of the physical models by distance. https://github.com/sunpy/sunkit-spex/pull/195

sunkit_spex/models/physical/thermal.py
samaloney: We can easily compose the models e.g.
```
line = LineEmission()
cont = ContinuumEmission()
thermal = line + cont
```
but the abundances and temp won't be linked an alternative is to use the them inside another model ignore the actual name and code but you get the idea

```
class Thermal(Model)

def __inti__(self, temp, em, abun):
    self.line = LineMission(temp, em, abun)
    self.cont = ...
    ....

def evaluate(self, energy, temp, em, abun)
    return self.line.evalute(...) + self.cont.evalute(...)
```

this might get the best of both?

sunkit_spex/models/physical/tests/test_thermal.py
```diff
@@ -49,62 +49,68 @@ def fvth_simple():
...
-        2.2103457e-21,
-        8.2594126e-22,
-        3.0853276e-22,
-        1.1521560e-22,
-    ] * SSW_INTENSITY_UNIT
-    return inputs, ssw_output
+    mg, al, si, s, ar, ca, fe = 8.15, 7.04, 8.1, 7.27, 6.58, 6.93, 8.1
+    inputs = (energy_edges, temperature, emission_measure, mg, al, si, s, ar, ca, fe, abundance_type)
+    inputs_class = (temperature, emission_measure, mg, al, si, s, ar, ca, fe, abundance_type)
+
```
samaloney, line 56: Might want to turn off autofarmatting for these you can use `# fmt: off` and `# fat: on` before and after block to allow you to format as you please maybe joining the line and wrapping at the line lenght 120 or something?

sunkit_spex/models/physical/thermal.py
jajmitchell: @KriSun95 thoughts on this?

sunkit_spex/models/physical/thermal.py
jajmitchell: I can see that this would streamline things, however the current way we have it essentially does do this, but with the advantage that `line_emission`, `continuum_emission` and `thermal_emission` can all be accessed and used with the scipy fitters as in the legacy code. For this reason I'd probably keep things broadly how they are, but of course can strip things out and streamline if that's the concensus

samaloney: > I can see that this would streamline things, however the current way we have it essentially does do this, but with the advantage that line_emission, continuum_emission and thermal_emission can all be accessed and used with the scipy fitters as in the legacy code.

I don't think that's important legacy stuff should be kept in legacy, should we even be exposing `line_emission`, `continuum_emission` and `thermal_emission`?

If you wanted to keep things as they are you could strip out a common base class and subclass or similar - in the future something will be updated in one model and not done in the other two places or vice versa.

jajmitchell: I now switched things over to be in line with your suggestion above. Let me know what you think and if its good I'll clean up the code to remove the commented out `thermal_emission` function

jajmitchell:
@samaloney can we just skip the py310 test?

samaloney:
We could but some what defeats the purpose no? Does it happen locally for you `tox -e oldestdeps` (assuming you have some version of python3.10 installed)

sunkit_spex/models/physical/thermal.py
```diff
@@ -197,58 +707,77 @@ def setup_default_abundances(filename=None):
...
+
+#     # print(continuum_flux.unit)
+#     # print(line_flux.unit)
+
+#     # flux = (continuum_flux + line_flux) * emission_measure
+
+#     # if (temperature.isscalar and emission_measure.isscalar) or (len(temperature) == 1 and len(emission_measure) == 1):
+#     #     flux = flux[0]
+
+#     # return flux
```
samaloney, line 765: Delete?

samaloney (approved):
Once the test fail is resolved this looks good 🚀

jajmitchell:
So the issue is that we need a `ValueError` message in `test_thermal.py` without specifying a match. However this won't pass the ruff checks during pre-commit:

`sunkit_spex/models/physical/tests/test_thermal.py:427:24: PT011 `pytest.raises(ValueError)` is too broad, set the `match` parameter or use a more specific exception
    |
426 | def test_energy_out_of_range_error():
427 |     with pytest.raises(ValueError):
    |                        ^^^^^^^^^^ PT011
428 |         thermal.ThermalEmission(6 * u.MK, 1e44 / u.cm**3, 8.15, 7.04, 8.1, 7.27, 6.58, 6.93, 8.1)([0.01, 10] * u.keV)
`
We require this behaviour as having a specific error message match here would break whenever the chianti files were changed. 

I have tried skipping this multiple ways including editing the `tox.ini` and the `pyproject.toml` file, however I can't get it to work, any thoughts?

KriSun95 (approved):
LGTM

Issue #51, opened 2021-06-17
Thick and Think target models updates for a 0.0.1 release


samaloney:
Discussion of a the minimal required updates for a release 

- [ ] Add units to inputs, calculation, and outputs see #46
- [ ] Refactor further away from IDL like implementation #21 and #45
- [ ] Documentation / examples

DanRyanIrish:
Regarding point 2, I think the bigger thing is to decide on the API.  Refactors for greater speed and/or Pythonic-ness can be done at any stage with little penalty so long as the API doesn't change.  I realise that these two issues are completely separable but if others agree, that will hopefully give more focus.  So the question is, what should the API be?

My initial feeling is that the models should start off as functions.  This is already the case.  Although I'd like to better understand the decision behind making the electron distribution a class.  I'm not suggesting this is not the right thing to do.  But we should understand whether it is the best choice.

samaloney:
The electron distribution functions need to be able to return the normalised integral as well as the original distribution and also return the continuous parts where applicable so that why I had them as classes but could work as functions too. It hard to know until someones tires it but something like this maybe

```python
def power_law(x, slope, norm, integral=False):
     if integral is False:
         return x.min(), x.max(), norm * x **slope
     elif integral is True
         return x.min(), x.max(), norm * x ** (slope - 1) # this isn't wright just the idea

def thick_target(photon_edges, edist_func, edist_params=None, observer_distance=1*au, integration_method=scipy.integragate.quad'):
   if not callable(edict_funct):
       raise Type('edist_func must be callable')
   for continuous_part in edict_funct(**edist_params, integral=True)
       lower_limits, upper_limits, values = continuous_part
       res = integration_method(.....)

flux = thick_target(photon_edges, power_law, edist_params={'slope': -5, 'norm': 1000})
```

PR #194, opened 2025-02-28
Adds Astropy wrapper to the Thick and Thin target models

jajmitchell:
Adds an astropy model class wrapper to the thick and thin target models.

This PR changes the format of the thick and thin target models by adding an Astropy class wrapper, ThickTarget and ThinTarget, which will be located in models/physical/nonthermal.py

Currently the model can be evaluated both alone and when combined with the albedo model. Below is a plot showing the results of evaluating the model both alone and with the albedo model:

Thicktarget_albedo

Need to address the following:

- [x] Add Astropy class wrapper to Thin target model.
- [x] Add observer distance in as a parameter.
- [x] Sort out unit handling so that fitting works.
- [x] Write tests.


sunkit_spex/models/physical/nonthermal.py
```diff
@@ -0,0 +1,1095 @@
...
+    "ThickTarget",
+    "ThinTarget",
+    "_get_integrand",
+    "_integrate_part",
+    "_split_and_integrate",
+    "bremsstrahlung_thick_target",
+    "bremsstrahlung_thin_target",
+    "collisional_loss",
+    "thick_fn",
+    "thin_fn",
```
samaloney, line 41: I don't think we necessarily want all these functions available to the user and documented.

Also could possibly just import these and only leave ether actual fitable models here e.g. `ThickTarget` and `ThinTarget`

sunkit_spex/models/physical/nonthermal.py
```diff
@@ -0,0 +1,1095 @@
...
+
+    q = Parameter(name="q", default=5, min=0.01, description="Slope above break", fixed=True)
+
+    eelow = Parameter(name="eelow", default=7, unit=u.keV, description="Low energy electron cut off", fixed=False)
+
+    eehigh = Parameter(name="eehigh", default=1500, unit=u.keV, description="High energy electron cut off", fixed=True)
+
+    total_eflux = Parameter(
+        name="total_eflux", default=1.5, unit=u.electron * u.s**-1, description="Total electron flux", fixed=True
+    )
```
samaloney, line 68: I think this is a good opportunity to use better names?

samaloney (changes requested):

sunkit_spex/models/physical/nonthermal.py
jajmitchell: Any thoughts on where we could locate them? Perhaps we could have a `_functions.py` for each set of model classes?

sunkit_spex/models/physical/nonthermal.py
jajmitchell: Yeah totally agreed, I will edit these tomorrow!

settwi:
hey this looks good so far!

i think the thin_target normalization might be different for the thick target one. it's hard to say, because there isn't an equivalent in the legacy `photon_models_for_fitting.py`, but the docstring makes it sound different:
https://github.com/sunpy/sunkit-spex/blob/8cc9076a2be361c506f860882017614ac02e46fb/sunkit_spex/legacy/emission.py#L683-L689

Trying out `bremsstrahlung_thin_target` and multiplying by 1e35 gives flux values which are way off, too:
```py
from sunkit_spex.legacy import emission

energies = np.geomspace(1, 200, 1000)
thin = emission.bremsstrahlung_thin_target(
    energies,
    # index below break
    p=2,
    # break energy
    eebrk=40,
    # index above break
    q=4,
    # low-E cutoff
    eelow=10,
    # high-E cutoff
    eehigh=500,
    # "electron flux density distribution"
    efd=True
)

flux = 3
thin *= flux * 1e35
thin.max()
# output: np.float64(3.018588658961549e-17)
```

as a final note, in a future PR, it would probably be fruitful to rewrite the nonthermal Bremsstrahlung stuff in a more pythonic way. they're basically just IDL ports

samaloney:
> as a final note, in a future PR, it would probably be fruitful to rewrite the nonthermal Bremsstrahlung stuff in a more pythonic way. they're basically just IDL ports
💯 see #51 for some initial thoughts

jajmitchell:
I am seeing the same issue as on the previous pull request with this specific warning, not sure exactly how to fix it, but apart from that, I think its good to go

samaloney (approved):

PR #195, opened 2025-03-04
Adds Distance and Constant multiplicative factor scaling models as Astropy models 

jajmitchell:
Adds an Distance and Constant multiplicative factor scaling models as Astropy models 

This PR adds `scaling.py` which includes `DistanceScale` and `Constant` which can be used alongside other models to scale for the observer distance or by an constant multiplicative factor, both the distance and the constant can be fit. 

distance

constant

- [x] Add tests

- [x] Add doc strings

### DIFF
--- a/changelog/195.feature.rst
+++ b/changelog/195.feature.rst
@@ -1,6 +1,1 @@
-Add new two new Astropy model classes
-
-:py:class:`sunkit_spex.models.scaling.DistanceScale`
-:py:class:`sunkit_spex.models.scaling.Constant`
-
-These classes can be used with physical models to scale the flux either by observer distance or by a constant multiplicative factor respectively.
+Add new two new Astropy model classes, :class:`sunkit_spex.models.scaling.DistanceScale` and :class:`sunkit_spex.models.scaling.Constant`. These classes can be used with physical models to scale the flux either by observer distance or by a constant multiplicative factor respectively.
--- a/changelog/195.feature.rst
+++ b/changelog/195.feature.rst
@@ -1,7 +1,6 @@
 Add new two new Astropy model classes
 
-:py:class:`sunkit_spex.models.scaling.DistanceScale` 
+:py:class:`sunkit_spex.models.scaling.DistanceScale`
 :py:class:`sunkit_spex.models.scaling.Constant`
 
-These classes can be used with physical models to scale the flux either by observer distance or by a constant multiplicative factor respectively. 
-
+These classes can be used with physical models to scale the flux either by observer distance or by a constant multiplicative factor respectively.
--- a/changelog/195.feature.rst
+++ b/changelog/195.feature.rst
@@ -1,0 +1,7 @@
+Add new two new Astropy model classes
+
+:py:class:`sunkit_spex.models.scaling.DistanceScale` 
+:py:class:`sunkit_spex.models.scaling.Constant`
+
+These classes can be used with physical models to scale the flux either by observer distance or by a constant multiplicative factor respectively. 
+
--- a/sunkit_spex/models/scaling.py
+++ b/sunkit_spex/models/scaling.py
@@ -20,9 +20,9 @@ class DistanceScale(FittableModel):
     _input_units_allow_dimensionless = True
 
     def evaluate(self, spectrum, observer_distance):
-        spectrum_distance_corrected = distance_correction(spectrum, observer_distance)
 
-        return spectrum_distance_corrected
+        return distance_correction(spectrum, observer_distance)
+
 
     def _parameter_units_for_data_units(self, inputs_unit, outputs_unit):
         return {"observer_distance": u.AU}
@@ -55,6 +55,4 @@ def distance_correction(spectrum, observer_distance):
 
     scale = AU_distance_cm**2 / observer_distance_cm**2
 
-    flux = spectrum * scale
-
-    return flux
+    return spectrum * scale
--- a/sunkit_spex/models/scaling.py
+++ b/sunkit_spex/models/scaling.py
@@ -1,0 +1,78 @@
+from functools import lru_cache
+
+import numpy as np
+from numpy.typing import NDArray
+from scipy.interpolate import RegularGridInterpolator
+from scipy.io import readsav
+
+import astropy.units as u
+from astropy.modeling import FittableModel, Parameter
+from astropy.units import Quantity
+
+from sunpy.data import cache
+
+__all__ = ["DistanceScale", "Constant"]
+
+
+class DistanceScale(FittableModel):
+
+    n_inputs = 1
+    n_outputs = 1
+    
+    observer_distance = Parameter(
+        name="observer_distance",
+        default=1,
+        unit=u.AU,
+        description="Distance to the observer in AU",
+        fixed=True,
+    )
+
+    _input_units_allow_dimensionless = True
+
+    # def __init__(self, *args, **kwargs):
+    #     self.energy_edges = kwargs.pop("energy_edges")
+
+    #     super().__init__(*args, **kwargs)
+
+    def evaluate(spectrum, observer_distance):
+
+        spectrum_distance_corrected = distance_correction(spectrum, observer_distance) 
+
+        return spectrum_distance_corrected
+
+    def _parameter_units_for_data_units(self, inputs_unit, outputs_unit):
+        return {"observer_distance": u.AU}
+
+
+
+class Constant(FittableModel):
+
+    n_inputs = 1
+    n_outputs = 1
+    
+    constant = Parameter(
+        name="constant",
+        default=1,
+        description="Multiplicative Constant",
+        fixed=True,
+    )
+
+    _input_units_allow_dimensionless = True
+
+    def evaluate(spectrum, constant):
+
+        return spectrum * constant
+
+
+
+def distance_correction(spectrum, observer_distance):
+
+    AU_distance_cm = 1*u.AU.to(u.cm)
+    observer_distance_cm = observer_distance.to(u.cm)
+
+    scale = AU_distance_cm**2 / observer_distance_cm**2
+
+    flux = spectrum * scale
+
+    return flux
+
--- a/sunkit_spex/models/scaling.py
+++ b/sunkit_spex/models/scaling.py
@@ -29,12 +29,7 @@ class DistanceScale(FittableModel):
 
     _input_units_allow_dimensionless = True
 
-    # def __init__(self, *args, **kwargs):
-    #     self.energy_edges = kwargs.pop("energy_edges")
-
-    #     super().__init__(*args, **kwargs)
-
-    def evaluate(spectrum, observer_distance):
+    def evaluate(self, spectrum, observer_distance):
 
         spectrum_distance_corrected = distance_correction(spectrum, observer_distance) 
 
@@ -42,8 +37,6 @@ class DistanceScale(FittableModel):
 
     def _parameter_units_for_data_units(self, inputs_unit, outputs_unit):
         return {"observer_distance": u.AU}
-
-
 
 class Constant(FittableModel):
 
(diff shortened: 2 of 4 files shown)


changelog/195.feature.rst
samaloney, line 1: So the way these change log entires get rendered in the end they should ideally be one a single line (no line break, new lines) even if multiple sentences. Also I _think_ you can just use `:class:`.

sunkit_spex/models/scaling.py
```diff
@@ -0,0 +1,56 @@
...
+    _input_units_allow_dimensionless = True
+
+    def evaluate(self, spectrum, observer_distance):
+        return distance_correction(spectrum, observer_distance)
+
+    def _parameter_units_for_data_units(self, inputs_unit, outputs_unit):
+        return {"observer_distance": u.AU}
+
+
+class Constant(FittableModel):
```
samaloney, line 29: There is already and astropy const model could that be used?

sunkit_spex/models/scaling.py
```diff
@@ -0,0 +1,56 @@
...
+    def evaluate(self, spectrum, constant):
+        return spectrum * constant
+
+
+def distance_correction(spectrum, observer_distance):
+    if isinstance(observer_distance, Quantity):
+        AU_distance_cm = 1 * u.AU.to(u.cm)
+        observer_distance_cm = observer_distance.to(u.cm)
+    else:
+        AU_distance_cm = 1.496e13
```
samaloney, line 51: Use astropy units e.g (1 *u.AU) then convert to cm at the end

sunkit_spex/models/scaling.py
```diff
@@ -0,0 +1,56 @@
...
+
+class DistanceScale(FittableModel):
+    n_inputs = 1
+    n_outputs = 1
+
+    observer_distance = Parameter(
+        name="observer_distance",
+        default=1,
+        unit=u.AU,
+        description="Distance to the observer in AU",
```
samaloney, line 16: It doesn't have to be in AU we can use astropy `length` equivalences so can take any lengths

sunkit_spex/models/scaling.py
```diff
@@ -0,0 +1,56 @@
...
+
+    def evaluate(self, spectrum, constant):
+        return spectrum * constant
+
+
+def distance_correction(spectrum, observer_distance):
+    if isinstance(observer_distance, Quantity):
+        AU_distance_cm = 1 * u.AU.to(u.cm)
+        observer_distance_cm = observer_distance.to(u.cm)
+    else:
```
samaloney, line 50: The more I see this the more I think we should just make the units mandatory on these kind or parameters

samaloney (changes requested):
First great to see all the PRs 🚀 

I think we need to be mindful about a few things:
1. Possible a more more descriptive name `InverseSquareScaling` or `NormalisedInverseSquareScaling`. With the later can could support other norm distance than 1AU
2. This can only work with a model which has output as some kind  of flux at 1AU but there are many models that don't how to we stop it being used in correctly - maybe just good docs but alway better to have some code check.

sunkit_spex/models/scaling.py
jajmitchell: So this is nothing to do with the values passed to the model, these will require units, this step ensures that the evaluate can operate both with and without units as the units are stripped whilst fitting and then reapplied to the result post fitting by the Astropy fitting routines.

sunkit_spex/models/scaling.py
jajmitchell: Ahh, I didn't realise, yeah we can probably just get rid of this one then

sunkit_spex/models/scaling.py
jajmitchell: Sounds good, I will see if we can keep the default as 1 AU, but enable the user to specify any length.

sunkit_spex/models/scaling.py
jajmitchell: Another thought on this, `Const1D` and `Const2D`, are slightly tricky for us as our input and output are of different dimensions, as we evaluate at energy edges, but our output are fluxes at energy centers . The `Constant` class that created here allows the output spectrum from our physical models to be piped through and scaled and therefore there is no issue with the dimensions. I think its a good model for us to keep in for this reason as it functions in the same way as the `DistanceScale` model, and is probably more useful for our specific use case

sunkit_spex/models/scaling.py
samaloney: But that should just work with the astropy fitter then e.g. they wrapped in this decorator https://github.com/astropy/astropy/blob/b7b050de7b43b0951c3c93570359fee493e5f532/astropy/modeling/fitting.py#L182

jajmitchell: The issue is that in our case evaluate acts as a wrapper to an external function, I think the fitter strips the units for evaluate, but then the external function called by evaluate must function with and without units, so that the model can be evaluated with units and also fit, where the unit stripping occurs

jajmitchell: Ahhh, I see, I didn't realise this was available, I'll add this in and make the relevant changes

KriSun95:
@jajmitchell, is it possible to update the image/examples at the top to show the changes since the PR was opened? I'm mainly thinking of showing the `|` changing to `*`.

jajmitchell:
Aside from the docs failure, I think that this is now ready for merging. 

`InverseSquareFluxScaling` now functions as a multiplicative model with correct unit handling.

`ScalarConstant` can be used with whatever operator desired. The example given is for use as a multiplicative constant. This model is much more simple than the Astropy constant models, and is more applicable to our usecase.

samaloney:
Need to merge or rebase `upstream main` into this branch to fix the doc builds

jajmitchell:
> Need to merge or rebase `upstream main` into this branch to fix the doc builds

I think this is sorted, but am unsure why the changelog is now failing?

sunkit_spex/models/scaling.py
```diff
@@ -0,0 +1,182 @@
...
+
+        if isinstance(observer_distance, Quantity):
+            return np.full(dimension, correction.value) * correction.unit
+        return np.full(dimension, correction) * (1 / u.cm**2)
+
+    def _parameter_units_for_data_units(self, inputs_unit, outputs_unit):
+        return {"observer_distance": u.AU}
+
+
+class ScalarConstant(FittableModel):
```
samaloney, line 95: Can a scaler not be constant? `Constant(Fitt...)`

sunkit_spex/models/scaling.py
```diff
@@ -0,0 +1,182 @@
...
+    """
+    ScalarConstant
+    model converts luminosity output of physical models to a distance scaled flux.
```
samaloney, line 98: This model is just a constand so is can be used to scale or add to a an existing model or on its own - I'd look at the doc string for the atrou Const1D model etc

sunkit_spex/models/scaling.py
```diff
@@ -0,0 +1,182 @@
...
+    Examples
+    ========
+    .. plot::
+        :include-source:
+
+        import astropy.units as u
+        import numpy as np
+        import matplotlib.pyplot as plt
+
+        from astropy.modeling.powerlaws import PowerLaw1D
```
samaloney, line 114: Not used?

sunkit_spex/models/scaling.py
```diff
@@ -0,0 +1,182 @@
...
+
+    """
+
+    n_inputs = 1
+    n_outputs = 1
+
+    constant = Parameter(
+        name="constant",
+        default=1,
+        description="Multiplicative Constant",
```
samaloney, line 159: Doesn't have to be multiplicative could be additive no?

sunkit_spex/models/scaling.py
```diff
@@ -0,0 +1,182 @@
...
+
+def distance_correction(observer_distance):
+    if isinstance(observer_distance, Quantity):
+        if observer_distance.unit.is_equivalent(u.AU):
+            observer_distance_cm = observer_distance.to(u.cm)
+        else:
+            raise ValueError("Observer distance input must be an Astropy length convertible to AU.")
+
+    else:
+        AU_distance_cm = 1 * u.AU.to(u.cm).value
```
samaloney, line 179:
```suggestion
        AU_distance_cm = 1 * u.AU.to_value(u.cm)
```

sunkit_spex/models/scaling.py
```diff
@@ -0,0 +1,182 @@
...
+
+    _input_units_allow_dimensionless = True
+
+    def evaluate(self, x, constant):
+        dimension = np.shape(x)[0] - 1
+
+        return np.full(dimension, constant)
+
+
+def distance_correction(observer_distance):
```
samaloney, line 171: Could this be in the evaluate of the model, what the reason advantage to separate?

samaloney (changes requested):
Nearly there I think the docs need some work. You need to add the new module to the docs and also the doc and doc strings and explanation could use a tweak or two.

sunkit_spex/models/scaling.py
jajmitchell: Agreed, given that its such a simple model its probably better in evaluate.

sunkit_spex/models/scaling.py
jajmitchell: apologies, I copied this from `InverseSquareFluxScaling` and forgot to edit it!

sunkit_spex/models/scaling.py
```diff
@@ -0,0 +1,182 @@
...
+        from astropy.modeling.powerlaws import PowerLaw1D
+        from astropy.visualization import quantity_support
+
+        from sunkit_spex.models.scaling import ScalarConstant
+        from astropy.modeling import Parameter, FittableModel
+
+        ph_energies = np.arange(4, 100, 0.5)
+        ph_energies_centers = ph_energies[:-1] + 0.5*np.diff(ph_energies)
+
+        class StraightLineModel(FittableModel):
```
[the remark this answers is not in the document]
jajmitchell: we need to edit the `StraightLineModel` and `GaussianModel` to output with length_output = length_input -1, I could quickly open a new PR to sort this out as its a very easy change?

sunkit_spex/models/scaling.py
jajmitchell: I just meant that its a Scalar constant in the sense its dimensionless. Do you think it'd be better to name it `Constant`?

sunkit_spex/models/scaling.py
```diff
@@ -0,0 +1,182 @@
...
+        from astropy.visualization import quantity_support
+
+
+        from sunkit_spex.models.scaling import InverseSquareFluxScaling
+        from astropy.modeling import Parameter, FittableModel
+
+        ph_energies = np.arange(4, 100, 0.5)
+        ph_energies_centers = ph_energies[:-1] + 0.5*np.diff(ph_energies)
+
+        class StraightLineModel(FittableModel):
```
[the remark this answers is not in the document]
jajmitchell: we need to edit the StraightLineModel and GaussianModel to output with length_output = length_input -1, I could quickly open a new PR to sort this out as its a very easy change?

jajmitchell:
@samaloney when you say "You need to add the new module to the docs", do you mean add relevant .rst files to the `/docs/api/` directory?

sunkit_spex/models/scaling.py
samaloney: But it doesn't have to be dimensionless does it, could have a units?

sunkit_spex/models/scaling.py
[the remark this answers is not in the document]
jajmitchell: They are used in some of the tests I think, but its just a case of adding in the new argument to those tests. I will do that on the PR that I have opened and hopefully it'll all run smoothly. This is a necessary change if we are to be able to use these as compound models with our physical models

sunkit_spex/models/scaling.py
jajmitchell: This is sorted now, I have rernamed as `Constant` and the model can now be evaluated and fit with any user specified units.

sunkit_spex/models/scaling.py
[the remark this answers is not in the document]
jajmitchell: Once this is sorted on the other PR, I think this should all function well

KriSun95 (approved):

samaloney (changes requested):
I think the docs configuration is missing? (Don't see the new functions in the build docs or change to docs config)

jajmitchell:
> I think the docs configuration is missing? (Don't see the new functions in the build docs or change to docs config)

I think this is now sorted

jajmitchell:
the docs fail, because the examples in the docstrings require the new `StraightLineModel` and `GaussianModel` to function. Should I just remove the examples then we can add them back in in a trivial PR once these two PRs are merged?

samaloney (approved):

PR #202, opened 2025-04-09
Enables `StraightLineModel` and `GaussianModel` to be evaluated at energy edges and return values at energy centers.

jajmitchell:
Enables `StraightLineModel` and `GaussianModel` to be evaluated at energy edges and return values at energy centers, by setting `photon_model` to `True`,  so that they can both be used alongside our physical models.


examples/fitting_simulated_data.py
```diff
@@ -43,8 +43,8 @@
 #
 # Let's start making a simulated photon spectrum
 
-sim_cont = {"slope": -1, "intercept": 100}
-sim_line = {"amplitude": 100, "mean": 30, "stddev": 2}
+sim_cont = {"photon_model": False, "slope": -1, "intercept": 100}
```
natsat0919, line 46: Maybe I'm overlooking something, but shouldn't these be photon models and so "photon_model" should be True?

jajmitchell: Not yet, as these examples use the srm etc, and so cannot yet be fit with units due to the astropy fitting issues, therefore its all done without units in the examples currently I think

sunkit_spex/models/models.py
```diff
@@ -2,27 +2,94 @@
...
+
+    input_units_equivalencies = {"keV": u.spectral()}
+
     slope = Parameter(default=1, description="Gradient of a straight line model.")
     intercept = Parameter(default=0, description="Y-intercept of a straight line model.")
 
-    @staticmethod
-    def evaluate(x, slope, intercept):
+    def __init__(self, *args, **kwargs):
+        self.photon_model = kwargs.pop("photon_model")
```
natsat0919, line 24: This is just a suggestion/idea: Since all our other physical models are photon models by default. Would it be worth it to have self.photon_model set to True by default. Unless user specifically includes "photon_model" if the params dictionary, then we either change it to False or keep it True? Just an idea, maybe it's an unnecessary feature

examples/fitting_simulated_data.py
natsat0919: ah right, that makes sense

sunkit_spex/models/models.py
jajmitchell: I have added this in, I think its a good plan. An error in case the input is not convertible to keV needs to be added in along with a corresponding test. I'll sort that tomorrow. The tests should still all work

natsat0919: Looks good!

sunkit_spex/models/models.py
```diff
@@ -2,27 +2,99 @@
...
+    _input_units_allow_dimensionless = True
+
+    input_units_equivalencies = {"keV": u.spectral()}
+
     slope = Parameter(default=1, description="Gradient of a straight line model.")
     intercept = Parameter(default=0, description="Y-intercept of a straight line model.")
 
-    @staticmethod
-    def evaluate(x, slope, intercept):
+    def __init__(self, slope=slope.default, intercept=intercept.default, photon_model=True, **kwargs):
```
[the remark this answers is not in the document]
jajmitchell: So the `SpectralAxis` should be populated by energy edges anyway right? If so, then these models will function the same as the physical models which are all evaluated at energy edges, and will do so as a default. I added the `photon_model` name in as, not only does this switch to energy edges but also provides the unit support to match the physical models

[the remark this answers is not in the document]
jajmitchell: @samaloney, we discussed the units during the last meeting. I should have mentioned, the reason we either need an initialisation argument for user specified output units, or to leave in the `photon_model` argument, is that when used  as an additive model in compound with other models such as our physical models, the output units must match, and the model can't see the y units just from passing the x units for evaluation

samaloney (approved):
LGTM - don't forget to create the issue and maybe mention this PR in it (about edges vs SpectralAxis etc)

Issue #205, opened 2025-05-01
Add albedo correction to legacy fitting


natsat0919:
### Describe the feature

Add the option to perform albedo correction using the old legacy code.

PR #206, opened 2025-05-01
Legacy albedo correction

natsat0919:
Referring to issue #205, this PR adds the option to perform albedo correction with the old legacy fitting code. The actual albedo code is based on what has been written for the new code (this is done in legacy/fitting/albedo.py), but the application of the albedo model is done in the fitter.py . Since I had to change some code in the core of the fitter, I would appreciate if others could try it out as well.

I'll be adding a STIX example notebook that will show how the correction is applied, but for now:

_fitter_object = Fitter(spec_file, srm_file)_

Define all the models...

add albedo with:
_fitter_object.albedo_corr = True_
_fitter_object.albedo_angle = 78.7 * u.deg #degrees. Whatever the angle is_
_fitter_object.albedo_anisotropy = 1 #this is the default value, you can remove the line if you don’t want to change the value._

By default .albedo_corr is false, so the fitting can be done the same way it has been done before without any changes to the user side of coding. 

Muriel and I tested it with OSPEX and the softwares agree well. For example for one fit, the OSPEX params are (thanks to Muriel for doing the fit):

EM1 = (0.481509  +/-  0.0964133) * 10^49
T1 = 19.3587  +/-        1.57184
EM2 = (0.0122828  +/-      0.0213339) * 10^49
T2 = 38.2509  +/-        11.9998
E_Flux = 3.69020  +/-        7.02093
delta = 4.73086  +/-       0.15462
E_Cut = 19.9307  +/-        10.6589

and the sunkit-spex fit with mcmc:

9058da51-6437-4c6d-93e3-95cd461391e8

The non-thermal params e-flux and e_cut with OSPEX have very large errors, hence the difference in the best-fit values. Nevertheless, they are still within the errors of each other. I'll update this PR once I get a plot with side by side comparison in of the OSPEX and sunkit-spex best-fit count models.


sunkit_spex/legacy/fitting/albedo.py
samaloney, line 1: Is this just a straight copy? Could you import the function that don't need changing rather than copying?

natsat0919: yes, as far as I remember, it should be. I didn't import it as I wasn't sure about the final format of the functions in the albedo model and I thought that legacy is a "stand alone" part of the package but I guess at this point I might as well import it. Saves the extra effort if we are make our own green matrices and update the functions

natsat0919: Actually now I remember, the way the energy edges are handled for astropy models doesn't align with the Fitter object so some of the array indexing doesn't work

settwi: would it be possible to write an adapter class or function between the two ? then most of the code wouldn’t need to get duplicated

samaloney: Yea I'm not sure but it looks like `_get_green_matrix` and `_calculate_albedo_matrix` are the same so these could be imported and then just define a different `get_albedo_matrix` if necessary?

natsat0919: @settwi Yeah that would be possible, but just wonder if it's worth the effort just for this. @samaloney, `_calculate_albedo_matrix ` is also different when calculating energy centres. Just the `_get_green_matrix` is an actual copy between the two. Wonder what's best to do; write a function/class that would adapt for the functions in the new code or just stick with the slightly adapted copied code?

settwi: one thing to consider is keeping the code [DRY](https://www.plutora.com/blog/understanding-the-dry-dont-repeat-yourself-principle)

basically if we repeat the code, debugging requires double the number of edits and tests need to be mirrored as well

settwi: frankly it looks like the `Albedo` model itself needs some more verification tests; at the moment it's just got ["smoke tests"](https://en.wikipedia.org/wiki/Smoke_testing_(software)) for the actual `source | Albedo` portion

settwi:
@natsat0919 @samaloney
plz check if the commit i pushed works? otherwise, we can reset it to a prior commit:
```bash
# Reset to commit just before last one
git reset --hard HEAD~1
# Force-push the branch back to that commit
git push --force
```

natsat0919:
@samaloney it should be up to date with main

KriSun95 (changes requested):
The comment I've left might not be an issue but it would be really good to check before merging otherwise it breaks a really important feature. I'm happy to approve if it all checks out.

sunkit_spex/legacy/fitting/fitter.py
```diff
@@ -2502,18 +2526,22 @@ def _calculate_model(self, **kwargs):
...
+                alebdo_excess_count[m][0] /= e
 
         self._energy_fitting_indices = _energy_fitting_indices_orig
         # numpy is better to store but all models need to be the same length, this might not be the case
         try:
-            return np.concatenate(tuple(mu))
+            return np.concatenate(tuple(mu)), np.concatenate(tuple(alebdo_excess_count))
+
         except ValueError:
             return [mu[i][0] for i in range(len(mu))]
```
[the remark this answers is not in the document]
natsat0919: That's a good point. I just assumed that because the examples were compiled correctly, it should work. I'll double check and let you know

PR #210, opened 2025-05-28
Fixes fitting by removing `input_unit_equivalencies` from thermal classes

jajmitchell:
## Fixes fitting by removing `input_unit_equivalencies` from thermal classes.

This is a very simple fix with only three lines changed and allows the thermal classes, `ThermalEmission`, `ContinuumEmission` and `LineEmission` to be used in compound fitting.

### DIFF
--- a/sunkit_spex/models/physical/thermal.py
+++ b/sunkit_spex/models/physical/thermal.py
@@ -157,7 +157,6 @@ class ThermalEmission(FittableModel):
 
     fe = Parameter(name="Fe", default=8.1, min=6.1, max=10.1, description="Fe relative abundance", fixed=True)
 
-    input_units_equivalencies = {"keV": u.spectral(), "K": u.temperature_energy()}
     _input_units_allow_dimensionless = True
 
     def __init__(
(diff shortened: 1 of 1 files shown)


sunkit_spex/models/scaling.py
```diff
@@ -74,15 +76,19 @@ def evaluate(self, x, observer_distance):
...
+            AU_distance_cm = (1 * u.AU).to_value(u.cm)
             observer_distance_cm = observer_distance * AU_distance_cm
 
         correction = 1 / (4 * np.pi * (observer_distance_cm**2))
         dimension = np.shape(x)[0] - 1
 
         if isinstance(observer_distance, Quantity):
             return np.full(dimension, correction.value) * correction.unit
-        return np.full(dimension, correction) * (1 / u.cm**2)
+        return np.full(dimension, correction)
```
samaloney, line 87: Cold this just be a single value with the correct shape?

samaloney (approved):

sunkit_spex/models/scaling.py
jajmitchell: I'm not sure I fully understand, isn't that essentially what np.full returns, an array of the correct shape populated by a single value?

samaloney: We only need a scalar value here and failing that a array with one element with the correct shape

```
import numpy as np
data = np.random.rand(24).reshape(2, 3, 4)
a = data/2
b = data/np.full(data.shape, 2) # create an array of same size as data for no reason
c = data/np.array(2).reshape(1, 1, 1) # if the constant has to be an array
a == b
c == b
```

KriSun95 (approved):

PR #212, opened 2025-06-05
Fix bug introduced during Albedo model refactor

samaloney:
Before the optimisation with caching the theta parameter was maintained as an angular quantity throughout. In order to use `lru_cache` some internal functions had to be changed to take python primitives (e.g float) but the code wasn't updated to account for this `np.cos(45 * u.deg)` -> `np.cos(np.deg2rad(45))`.

### DIFF
--- a/sunkit_spex/models/physical/albedo.py
+++ b/sunkit_spex/models/physical/albedo.py
@@ -215,10 +215,10 @@ def get_albedo_matrix(energy_edges: Quantity[u.keV], theta: Quantity[u.deg], ani
     >>> e = np.linspace(5,  500, 5)*u.keV
     >>> albedo_matrix = get_albedo_matrix(e,theta=45*u.deg)
     >>> albedo_matrix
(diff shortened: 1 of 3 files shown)


natsat0919 (approved):
This fixes the problem. Now the sunkit-spex albedo corrections agree with OSPEX. I cross-checked this through multiple spectral fits and all agree well.

samaloney:
Good catch on the angle in the if statements @natsat0919. I know the IDL code sets some array values to 0 but I don't really see the benefit so I would argue we shouldn't do this specific step.

sunkit_spex/models/physical/albedo.py
```diff
@@ -182,6 +182,7 @@ def _calculate_albedo_matrix(energy_edges: tuple[float], theta: float, anisotrop
     X, Y = np.meshgrid(energy_centers, energy_centers)
 
     albedo_interp = albedo_interpolator((X, Y))
+    albedo_interp[albedo_interp < 1e-10] = 0
```
samaloney, line 185:
```suggestion
```

PR #213, opened 2025-06-10
Fix stix loader to calculate count rate correctly

jajmitchell:
This PR changes the calculation in the STIX loader to calculate count rate correctly.

We calculate the mean rather than summed count rate for selected time range and adjust the errors accordingly.


jajmitchell:
I accidentally branched from an open PR and so am closing and opening a clean PR. Apologies!

PR #214, opened 2025-06-10
Calculates mean rather than summed count_rate

jajmitchell:
## This PR fixes a bug in `STIXLoader` that incorrectly calculates the average count rate for a given time range

Calculate mean rather than summed count rate for selected time interval and correct error calculation accordingly.


KriSun95 (approved):
LGTM

natsat0919:
well spotted!

natsat0919 (approved):

Issue #188, opened 2025-02-13
Add STIX fitting notebook example


natsat0919:
### Describe the feature

We need an example notebook for STIX (like we have for RHESSI and NuSTAR)

PR #217, opened 2025-06-24
Add STIX example notebook

natsat0919:
An example notebook for fitting STIX spectra (issue https://github.com/sunpy/sunkit-spex/issues/188). This shows an example of fitting a single STIX spectrum as well as a joint fit with STIX imaging and background detectors when an attenuator is used.

Note: The STIX joint fit is quiet complex and takes around 2 minute to fit which slows down the example notebook build.


settwi:
Hey Natalia this is looking good. I'm able to build the docs locally and the fits look to have converged. It is rather strange that the joint fit is so slow, even without running MCMC. Are there redundant operations happening somewhere, or unnecessary reading/writing to files? That could slow things down. I am not sure how the internals of the joint fits work.

Here are a few notes to consider:
- The runtime, deprecation, etc. warnings shouldn't be suppressed. If we keep suppressing them and Numpy deprecates something, all of the test will break and we won't know why.
https://github.com/sunpy/sunkit-spex/blob/b91707239beb3ccb9ff8147ed343d8be657d3683/examples/fitting_STIX_spectra.py#L54-L58

- For setting plotting parameters, it would be a better approach to use a matplotlib style sheet, `matplotlib.pyplot.rc_context`, or to update `matplotlib.rcParams` directly, rather than switching back and forth between predefined sizes by hand
https://github.com/sunpy/sunkit-spex/blob/b91707239beb3ccb9ff8147ed343d8be657d3683/examples/fitting_STIX_spectra.py#L80-L88

- For the following warning, the user shouldn't be told to ignore it. If we want them to ignore it, then it should be suppressed.
https://github.com/sunpy/sunkit-spex/blob/b91707239beb3ccb9ff8147ed343d8be657d3683/sunkit_spex/extern/stix.py#L506-L512

- For printing fit parameters, just use `fitter.show_params` which will format better in notebooks and HTML than just `params`
https://github.com/sunpy/sunkit-spex/blob/b91707239beb3ccb9ff8147ed343d8be657d3683/examples/fitting_STIX_spectra.py#L191-L193

- Along similar lines, we should use [Markdown tables](https://www.codecademy.com/resources/docs/markdown/tables) when possible so they render in the web docs and notebooks nicely, as opposed to .rst tables
https://github.com/sunpy/sunkit-spex/blob/b91707239beb3ccb9ff8147ed343d8be657d3683/examples/fitting_STIX_spectra.py#L248-L272

natsat0919:
@settwi thanks for the comments!

> * For setting plotting parameters, it would be a better approach to use a matplotlib style sheet, `matplotlib.pyplot.rc_context`, or to update `matplotlib.rcParams` directly, rather than switching back and forth between predefined sizes by hand

using layout="contrained".... doesn't work well for the multi-panel fits. Similarly, updating the plot size with rcParams would mean I would have to do that before a joint it and then also a single fit. I can't see what would be the benefit of it be rather than defining the figure size like it's done here. Maybe I'm missing something here.

> * For printing fit parameters, just use fitter.show_params which will format better in notebooks and HTML than just params

I chose to use that instead of the .show_params is it formats the best-fit values and errors a lot nicer than .show_params. the "e" notation in show_params is less nice imo and only shows 3 significant figures. But I'm okay with using show_params if there is a strong preference for that.

> * For the following warning, the user shouldn't be told to ignore it. If we want them to ignore it, then it should be suppressed.

The error would be raised during a joint fit even if the time range doesn't cover an attenuation state. I think that trying to make the STIX loader aware if the fit is joint or not is a lot more effort than simply warning the user that this error might be raised. I think it's clear from the message that this could be a possible reason for such an error message. If you have time to edit the code to overcome this, feel free to edit the loader.

> * Along similar lines, we should use [Markdown tables](https://www.codecademy.com/resources/docs/markdown/tables) when possible so they render in the web docs and notebooks nicely, as opposed to .rst tables

I think I'm a bit confused by this. The table in the python file is using this notation.

settwi:
> I chose to use that instead of the .show_params is it formats the best-fit values and errors a lot nicer than .show_params. the "e" notation in show_params is less nice imo and only shows 3 significant figures. But I'm okay with using show_params if there is a strong preference for that.

Printing is fine. Do you trust more than 3 significant figures? ;)

> using layout="contrained".... doesn't work well for the multi-panel fits. Similarly, updating the plot size with rcParams would mean I would have to do that before a joint it and then also a single fit. I can't see what would be the benefit of it be rather than defining the figure size like it's done here. Maybe I'm missing something here.

The benefit of using the `with plt.rc_context` construction is that it is more Pythonic and doesn't change the global state of matplotlib. You can set one dict of params per style you want and use that. It's harder to mess up because you don't have to set/unset the styles each time. Also, you don't need to use constrained layout, tight is fine.

> I think I'm a bit confused by this. The table in the python file is using this notation.

If you click "download notebook" from the docs page, the tables render incorrectly in Jupyter. I don't think this is a problem limited to this notebook, I've noticed it elsewhere too but I wanted to point it out here.

image

> The error would be raised during a joint fit even if the time range doesn't cover an attenuation state. I think that trying to make the STIX loader aware if the fit is joint or not is a lot more effort than simply warning the user that this error might be raised. I think it's clear from the message that this could be a possible reason for such an error message. If you have time to edit the code to overcome this, feel free to edit the loader.

Putting a warning and then saying "you can ignore this if you want" goes against the philosophy of giving a warning. It's like ignoring the numpy warnings at the top of the file. It's a makeshift solution to the real issue which I wrote a little more about below.

The `STIXLoader.select_time` method should not be setting the interval as it does. Looking in the example gallery, you set the times to `"2024-10-01T22:10:10", "2024-10-01T22:10:18"` but they get updated to `2024-10-01T22:10:10.000, 2024-10-01T22:23:52.774`? This even happens when you only have one spectrum, so I don't think this has anything to do with the multi-spectrum fitting:
https://github.com/natsat0919/sunxspex/blob/d67e968f19b0a83d8e33c105a0920efdba306c02/examples/fitting_STIX_spectra.py#L113-L114

The RHESSI loader has some modifications so that warning is not emitted

https://github.com/natsat0919/sunxspex/blob/d67e968f19b0a83d8e33c105a0920efdba306c02/sunkit_spex/extern/rhessi.py#L267-L275

Or if you do try to use the old style update, it gives you a deprecation warning

https://github.com/natsat0919/sunxspex/blob/d67e968f19b0a83d8e33c105a0920efdba306c02/sunkit_spex/extern/rhessi.py#L601-L621

https://github.com/natsat0919/sunxspex/blob/d67e968f19b0a83d8e33c105a0920efdba306c02/sunkit_spex/extern/rhessi.py#L840-L845

natsat0919:
> The benefit of using the `with plt.rc_context` construction is that it is more Pythonic and doesn't change the global state of matplotlib. You can set one dict of params per style you want and use that. It's harder to mess up because you don't have to set/unset the styles each time. Also, you don't need to use constrained layout, tight is fine.

The only thing that's set in the examples for every plots is just the figure size, I think that's pretty straightforward to define. I tried different layouts but they don't work for joint fits because the text overflows and it makes the plots too small. Anyways, I changed all the other non-joint fitting plots to take layout=tight instead of defining the fig size.

> If you click "download notebook" from the docs page, the tables render incorrectly in Jupyter. I don't think this is a problem limited to this notebook, I've noticed it elsewhere too but I wanted to point it out here.

I'm not sure what's wrong there. When I try to use the format that works with jupyter notebook markdown then it doesn't compile correctly to the html file. I'll keep it as is for now as the priority is for the html to be compiled correctly and raise an issue for the future.

> The error would be raised during a joint fit even if the time range doesn't cover an attenuation state. I think that trying to make the STIX loader aware if the fit is joint or not is a lot more effort than simply warning the user that this error might be raised. I think it's clear from the message that this could be a possible reason for such an error message. If you have time to edit the code to overcome this, feel free to edit the loader.

I checked the code again and it's just a bug that propagated through the code as the `select_time` method updated start and end time separately. Hence the error was raised before the end time was updated. I also didn't realise the RHESSI time loading methods have changed so dramatically. It's not a good idea to use different time loading methods between the different instrument loaders. So I updated the STIX loader to use `update_event_time` and `update_background_time` methods instead of `select_time` to match the RHESSI loader.

natsat0919:
I didn't commit it here just now, I have it fixed locally. But I can do that now. Feel free to check/edit if I forgot something.

examples/fitting_STIX_spectra.py
```diff
@@ -86,7 +95,7 @@
 # Since the STIX spectrum is the only one loaded it is under the `"spectrum1"` entry.
 #
 # Default energy range plotted is all energies but the user can define an energy rangem or ranges.
-# Ranges are inclusive at the bounds and here we see the 4$-$15 keV, 15$-$30 keV, and 30$-$60 keV ranges.
+# Ranges are inclusive at the bounds and here we see the 4-15 keV, 15-30 keV, and 30-60 keV ranges.
```
samaloney, line 98: If you really want math mode need to use ``:math:`4-14` keV`` or similar

examples/fitting_STIX_spectra.py
```diff
@@ -0,0 +1,388 @@
...
+# | Model Parameter            |Recent OSPEX Fit            | This Work (MCMC, not shown)      | This Work (normal fit)     |
+# +============================+============================+==================================+============================+
+# | T [MK]                     | 19.61 |pm| 1.29            | 19.07 (-0.35, +0.26)             | 18.64 |pm| 0.22            |
+# +----------------------------+----------------------------+----------------------------------+----------------------------+
+# | EM [cm |-3| ]              | 471.90 |pm| 91.64 |x| |46| | 484.15 (-17.55, +23.02) |x| |46| | 554.37 |pm| 20.40 |x| |46| |
+# +----------------------------+----------------------------+----------------------------------+----------------------------+
+# | Superhot T [MK]            | 42.36 |pm| 8.16            | 37.57 (-2.12, +1.61)             | 39.58 |pm| 0.96            |
+# +----------------------------+----------------------------+----------------------------------+----------------------------+
+# | Supehot EM [cm |-3| ]      | 7.70 |pm| 6.61 |x| |46|    | 13.70 (-2.46, +4.59) |x| |46|    | 13.55 |pm| 1.46 |x| |46|   |
```
samaloney, line 245:
```suggestion
# | Superhot EM [cm |-3| ]      | 7.70 |pm| 6.61 |x| |46|    | 13.70 (-2.46, +4.59) |x| |46|    | 13.55 |pm| 1.46 |x| |46|   |
```

samaloney (approved):
Looks good to me. I'am curious as to why the factor between the imaging and bkg detector is ~80.

PR #221, opened 2025-07-05
Cleanup root

settwi:
A few things
1. Remove unused `requirements.txt`
2. Add `tox` to the `tests` optional dependencies
3. In `pyproject.toml`, combine `tests` and `docs` into `dev`--the docs serve as part of the tests, so they really shouldn't be separate.

### DIFF
--- a/changelog/221.trivial.rst
+++ b/changelog/221.trivial.rst
@@ -1,0 +1,1 @@
+Remove unused files from root directory; add `dev` target to pyproject


Cadair:
The main reason for not having tox be a test dep is that it isn't one :stuck_out_tongue: you don't need tox to run pytest, and tox installs the tests deps to run pytest.

I think adding tox as a dev dep would make sense (I'd be up for doing that in the template), but not a test dep or you get tox installing tox which isn't needed.

settwi:
Sweet, thanks for the suggestions folks!

samaloney (approved):
LGTM

PR #227, opened 2025-08-04
Adds SRM model with unit handling

jajmitchell:
## SRM model handles units

Adds in the SRM model developed in the PR [https://github.com/sunpy/sunkit-spex/pull/163](url). 

This enables physical models to be piped through the SRM model with correct unit handling and then fit to data in count space.


jajmitchell:
I have read through the co-pilot suggestions, I think it makes sense to merge as I am going to work immediately on the fitter which will require a different approach of unit handling and testing anyway

samaloney (approved):

examples/fitting_simulated_data.py
```diff
@@ -150,47 +182,86 @@
 #
 # Try and ensure we start fresh with new model definitions
 
-ph_mod_4astropyfit = StraightLineModel(**guess_cont) + GaussianModel(**guess_line)
-count_model_4astropyfit = (ph_mod_4fit | srm_model) + GaussianModel(**guess_gauss)
+# guess_cont = {"slope": -0.5 * u.ph / u.keV**2, "intercept": 80 * u.ph / u.keV}
```
settwi, line 186: @jajmitchell i'm taking a first pass through this PR. i think we need to remove commented blocks if they aren't part of the actual example

sunkit_spex/models/instrument_response.py
```diff
@@ -1,15 +1,46 @@
...
 
 class MatrixModel(Fittable1DModel):
-    def __init__(self, matrix):
-        self.matrix = Parameter(default=matrix, description="The matrix with which to multiply the input.", fixed=True)
-        super().__init__()
+    name = "SRM"
+
+    c = Parameter(fixed=True)
+
+    def __init__(self, matrix, input_axis, output_axis, input_spec_units, output_spec_units, c):
```
settwi: i think this could benefit from type annotations

sunkit_spex/models/instrument_response.py
```diff
@@ -1,15 +1,46 @@
...
+        self.input_spec_units = input_spec_units
+        self.output_spec_units = output_spec_units
+        self.input_axis = input_axis
+        self.output_axis = output_axis
+        self.matrix = matrix
+        super().__init__(c)
```
settwi: what is "c" ? i'm surprised Ruff didn't complain about this

sunkit_spex/models/instrument_response.py
```diff
@@ -1,15 +1,46 @@
...
 
-    def evaluate(self, model_y):
+    _input_units_allow_dimensionless = True
+
+    def evaluate(self, x, c):
         # Requires input must have a specific dimensionality
```
settwi: Typo

settwi: `# Requires input to have a specific dimensionality`

settwi:
hey @jajmitchell i'm taking a look through this, i started a review and will request a couple minor changes but i think it looks good overall. i might just push some commits up fixing typos if that is ok

settwi:
does [f38afa1](https://github.com/sunpy/sunkit-spex/pull/227/commits/f38afa13eab7350f551a10625284fbfb344ef62d) really remove the example?

sunkit_spex/models/instrument_response.py
```diff
@@ -1,15 +1,46 @@
...
+    @property
+    def input_units(self):
+        return self.input_spec_units
+
+    @property
+    def return_units(self):
+        return self.output_spec_units
+
+    def _parameter_units_for_data_units(self):
+        return {"c": self.output_spec_units["y"] / self.input_spec_units["x"]}
```
settwi: I am not a fan of calling the "input_spec_units" keys `x` and the "output_spec_units` keys `y`. they are not descriptive.

Also, why don't we just pass it as a single parameter rather than a dict?

examples/fitting_simulated_data.py
```diff
@@ -37,87 +40,110 @@
...
 #####################################################
 #
 # Add some noise
 np_rand = np.random.default_rng(seed=10)
-sim_count_model_wn = sim_count_model + (2 * np_rand.random(sim_count_model.size) - 1) * np.sqrt(sim_count_model)
+sim_count_model_wn = (
+    sim_count_model + (2 * np_rand.random(sim_count_model.size)) * np.sqrt(sim_count_model.value) * u.ct / u.keV
+)
+
+obs_spec = Spectrum(sim_count_model_wn.reshape(-1), spectral_axis=ph_energies)
```
settwi, line 116: @jajmitchell this is not used after plotting, especially now since the astropy model is commented out.

i think this example needs to get reworked from the ground up to remove stray bits like this. if i was a user and saw this object created then ignored, i would be rather confused. i was actually just confused looking through this example

sunkit_spex/models/instrument_response.py
```diff
@@ -1,15 +1,46 @@
...
-        return model_y @ self.matrix
+
+        input_widths = np.diff(self.input_axis)
+        output_widths = np.diff(self.output_axis)
+
+        flux = ((x * input_widths) @ self.matrix * c) / output_widths
+
+        if hasattr(c, "unit"):
+            return flux
+        return flux.value
```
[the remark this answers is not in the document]
settwi: i agree with the Copilot note

sunkit_spex/models/instrument_response.py
settwi: ^ Astropy convention

examples/fitting_simulated_data.py
```diff
@@ -37,87 +40,110 @@
...
-
-plt.text(80, 170, "(ph_model(sl,in,am1,mn1,sd1) | srm)", ha="right", c="tab:blue", weight="bold")
-plt.text(80, 150, "+ Gaussian(am2,mn2,sd2)", ha="right", c="tab:orange", weight="bold")
-plt.show()
+with quantity_support():
+    plt.figure()
+    plt.plot(ph_energies_centers, (ph_model | srm_model)(ph_energies), label="photon model features")
+    plt.plot(ph_energies_centers, GaussianModel(**sim_gauss)(ph_energies), label="gaussian feature")
+    plt.plot(ph_energies_centers, sim_count_model, label="total sim. spectrum")
+    plt.plot(obs_spec._spectral_axis, obs_spec.data, label="total sim. spectrum + noise", lw=0.5)
```
settwi, line 128: These should be plotted using `plt.stairs` to show that the X-ray spectra are binned

examples/fitting_simulated_data.py
```diff
@@ -150,47 +182,86 @@
...
+plt.figure(layout="constrained")
+
 row_labels = (
-    tuple(sim_cont)[-2:] + tuple(f"{p}1" for p in tuple(sim_line)[-3:]) + tuple(f"{p}2" for p in tuple(sim_gauss)[-3:])
-)
-column_labels = ("True Values", "Guess Values", "Scipy Fit", "Astropy Fit")
-true_vals = np.array(tuple(sim_cont.values())[-2:] + tuple(sim_line.values())[-3:] + tuple(sim_gauss.values())[-3:])
-guess_vals = np.array(
-    tuple(guess_cont.values())[-2:] + tuple(guess_line.values())[-3:] + tuple(guess_gauss.values())[-3:]
+    tuple(sim_cont) + tuple(f"{p}1" for p in tuple(sim_line)) + ("C",) + tuple(f"{p}2" for p in tuple(sim_gauss))
```
settwi, line 254: The scipy method does not respect the `fixed=True` status on the photon/ct scaling parameter

examples/fitting_simulated_data.py
KriSun95: I would go further and say any models should use `plt.stairs` but the final data `plt.plot` should use `plt.errorbar`.

sunkit_spex/models/instrument_response.py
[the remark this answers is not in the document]
KriSun95: I wrote a similar comment later then found this. I agree too.

sunkit_spex/models/instrument_response.py
KriSun95: I think, I think, `c` is a stand in for when the matrix model will have response parameters for gain slopes and offsets. The other reason was because Astropy complains when models don't have a fittable parameter. @jajmitchell can correct me on this.

KriSun95 (changes requested):

sunkit_spex/models/instrument_response.py
```diff
@@ -1,15 +1,46 @@
...
+    _input_units_allow_dimensionless = True
+
+    def evaluate(self, x, c):
         # Requires input must have a specific dimensionality
-        return model_y @ self.matrix
+
+        input_widths = np.diff(self.input_axis)
+        output_widths = np.diff(self.output_axis)
+
+        flux = ((x * input_widths) @ self.matrix * c) / output_widths
```
KriSun95, line 31: Just looking at the maths here, I think the `/output_widths` here is a typo. This would likely move the returned flux away from the data units, would it not?

sunkit_spex/models/instrument_response.py
samaloney: Yea it's the way get the units + model to work with astropy

jajmitchell:
@settwi @KriSun95, thanks for the comments, I will make some changes based on the suggestions, however perhaps at this stage its best to remove the `fitting_simulated_data.py` example, as currently the Astropy models can't be used for fitting so this will all dramatically change, unless you think we should include an example fitting simulated data with the legacy code? If so, that should probably be added in a separate pr and the current example removed in this one.

settwi:
> @settwi @KriSun95, thanks for the comments, I will make some changes based on the suggestions, however perhaps at this stage its best to remove the `fitting_simulated_data.py` example, as currently the Astropy models can't be used for fitting so this will all dramatically change, unless you think we should include an example fitting simulated data with the legacy code? If so, that should probably be added in a separate pr and the current example removed in this one.

Sure, good idea.

Feel free to revert the matrix commit I made btw. I’m not sure if it lines up with the intention of the “c” parameter. Making it transparent (you don’t need to pass it explicitly to the constructor) seemed like a better approach to me

jajmitchell:
@settwi @samaloney @KriSun95  would it make sense to sort this and merge it, or given that the unit handling will completely change within the next week or two with the implementation of the fitter, should we close this PR and add the SRM in the fitting PR, along with the changes to the unit handling of the other models?

samaloney:
I don't have a strong opinion either way, but prob lean towards merging. Also there's now bunch of changes to fix the doc builds so if we don't merge this I'll cherry pick then to a separate PR.

KriSun95:
Sure! I'm happy to vote to merge then decide on dimension handling stuff later.

When it comes to the `fitting_simulated_data.py`, can we keep it and just show the fit with Scipy `minimize`? I know the Astropy fit doesn't work but I still think it's useful to see a full fit from start to finish with what the newer code is capable of just now.

A simulated/custom data fit example for the legacy code already exists in the [Fitting Custom Spectra](https://sunkit-spex.readthedocs.io/en/latest/generated/gallery/fitting_custom_spectra.html) example.

PR #231, opened 2025-08-26
Thermal brems bugfixes

### LINKED ISSUE #229: Legacy `thermal._calculate_abundances` modifies global state inappropriately
The default abundances in `sunkit_spex.legacy.thermal._calculate_abundances` are fetched from the module-level "constant" (not actually constant, as will become clear shortly) `DEFAULT_ABUNDANCES`. Then, the relative abundance array is multiplied onto the abundances. This modifies the state of the `thermal` module, making it impossible to properly fit the abundances without a hack.

https://github.com/sunpy/sunkit-spex/blob/835592215724074dcafddaa4193b4f77866b11f1/sunkit_spex/legacy/thermal.py#L766-L787

Line 766 shows the assignment. Line 786 shows the operation which modifies the module variable state.

This can be reproduces as follows
```py
from sunkit_spex.legacy import thermal
data = thermal.DEFAULT_ABUNDANCES[thermal.DEFAULT_ABUNDANCE_TYPE].data
print(data)
# Prints: [1.00000000e+00 7.94327557e-02 4.36516170e-11 8.70964689e-11 ...

data += 1
print(thermal.DEFAULT_ABUNDANCES[thermal.DEFAULT_ABUNDANCE_TYPE].data)
# Prints: [2.        1.0794327 1.        1.        1.        1.0003891 1.0001 ...
```

I'm shocked this wasn't caught during testing. I hope the new modules do not have this issue.

### LINKED ISSUE #230: Please delete dead code
https://github.com/sunpy/sunkit-spex/blob/835592215724074dcafddaa4193b4f77866b11f1/sunkit_spex/models/physical/thermal.py#L1336-L1613

settwi:
Fixes #229 for legacy and dev't sections of code

### DIFF
--- a/sunkit_spex/legacy/thermal.py
+++ b/sunkit_spex/legacy/thermal.py
@@ -321,7 +321,7 @@ def _continuum_emission(energy_edges_keV, temperature_K, abundances):
     temperature_K: 1-D array-like
         The temperature(s) of the plasma in unit of K.  Must not be a scalar.
 
-    abundances: 1-D `numpy.array` of same length a DEFAULT_ABUNDANCES.
+    abundances: 1-D `numpy.array` of same length as DEFAULT_ABUNDANCES.
         The abundances for the all the elements.
     """
     # Handle inputs and derive some useful parameters from them
@@ -411,7 +411,7 @@ def _line_emission(energy_edges_keV, temperature_K, abundances):
     temperature_K: 1-D array-like
         The temperature(s) of the plasma in unit of K.  Must not be a scalar.
 
-    abundances: 1-D `numpy.array` of same length a DEFAULT_ABUNDANCES.
+    abundances: 1-D `numpy.array` of same length as DEFAULT_ABUNDANCES.
         The abundances for the all the elements.
     """
     n_energy_bins = len(energy_edges_keV) - 1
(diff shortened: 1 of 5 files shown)


samaloney:
Closes #229

sunkit_spex/legacy/thermal.py
```diff
@@ -764,15 +768,15 @@ def _warn_if_input_outside_valid_range(input_values, grid_range, param_name, par
 
 
 def _calculate_abundances(abundance_type, relative_abundances):
-    abundances = DEFAULT_ABUNDANCES[abundance_type].data
+    abundances = fetch_default_abundances(abundance_type)
```
samaloney: What about then we don't need a new function or to alter the names?
```suggestion
    abundances = DEFAULT_ABUNDANCES(abundance_type).data.copy()
```

samaloney (changes requested):
Good catch! What would you think about keeping original name and just do a `.copy()` inside the functions?

Also might be good to move the removal of the other code to a different PR - I'm not sure we ever fully decided on which approach to use and why so should make it easier to find.

Also a test for the new code to make sure this doesn't happen again would be good.

sunkit_spex/legacy/thermal.py
settwi: sure we could do a copy. the reasoning was, then whenever you reference DEFAULT_ABUNDANCES, you need to remember to do the .copy. with the function it does it by default

settwi:
Re-running `tox` locally and then i'll update. oops

settwi:
@samaloney i tried to keep the commit history clean for this one. maybe we could do a merge commit?

samaloney (approved):

KriSun95 (approved):
LGTM

PR #239, opened 2025-11-20
Fix spectrum object

samaloney:
This PR was originally supposed to fix slicing and arithmetic operations, and others things that were not working as expected which it does but not as trivially as before because in the process found some other deeper issues.

The `Spectrum` object now **requires** a `spectral_axis` to be given in all cases as either a `Quantity` or `SpectralAxis`. So the minimum requirement to create a `Spectrum` is now a data array and corresponding spectral axis (matching at least one dimension of the data). If a `WCS` is also given we try to ensure the given `spectral_axis ` "matches" the given `WCS` at lest in terms of the bin centers. In doing some of this it became obvious I don't have a clear idea of what the API should be some of the could maybe be made less convoluted with sunpy/ndcube/pull/713

In the interest of expediency and I think we should limit the API we can always expand it later much harder to reduce but we should discuss this.

Still some code be removed because we now require a `spectral_axis` and also edge cases e.g if you slice away the spectral_axis what should happen - get a NDCube, error ...

There is currently some code to automatically find the `spectral_axis_index` which is nice but again for the moment we could also require that as would simplify the code?

Also the `__init__` is mess hard to follow and sometime has self referencing calls I think could be refactored to be more linear and straightforward but don't want to do that until we have some discussion agreement


KriSun95:
The PR adding the spectrum object looks really good. I think it would be better to move anything adding to the ability of the spectrum object to a future PR. 

My reasoning is that the main building block for the fitting will be just having a single energy axis. To me, anything else moves into either a specific case for the fitting or into the instrument data container.

I've been playing around with the spectrum object for any given count spectrum and it's really nice. I'm just now looking at adding uncertainties and using a mask on the bins too.

samaloney:
Hey yea I think I can get this in to mergeable state but I would like to try to get the other examples working as these all share the same energy axis. 

I'd also like get some feedback from the NDCube experts @Cadair and @DanRyanIrish on the approach thus far. For example an alternative approach for the 2d example with data from different detectors would be to create a `SpectrogramSequnce` build on `NDCubeSequence`?

jajmitchell:
Ignore my previous comment, I think I've worked it out

sunkit_spex/spectrum/spectrum.py
```diff
@@ -1,57 +1,200 @@
...
+    forward_transform.inverse = SpectralTabular1D(array, lookup_table=np.arange(len(array)) * u.pix)
 
-    return SpectralGWCS(forward_transform=forward_transform, input_frame=coord_frame, output_frame=spec_frame)
+    tabular_gwcs = SpectralGWCS(
+        original_unit=orig_array.unit,
+        forward_transform=forward_transform,
+        input_frame=detector_frame,
+        output_frame=output_frame,
+    )
+    tabular_gwcs.bounding_box = None
```
samaloney, line 188: If I remove this everything breaks but I don't really understand why, I think it units somewhere but not sure.

sunkit_spex/spectrum/spectrum.py
```diff
@@ -1,57 +1,200 @@
...
+        """
+        Return a deep copy of the object.
+
+        Convenience method so user doesn't have to import the
+        :mod:`copy` stdlib module.
+        """
+        return copy.deepcopy(self)
+
+
+def gwcs_from_array(array, flux_shape, spectral_axis_index=None):
```
samaloney, line 55: Modified version of what in specutils - not sure if this is the best approach.

jajmitchell:
@samaloney Is the intention to store the srm matrix and photon axis in its own NDcube? Or is it possible to store this in the spectrum object and have the axes sliced when the count axis is sliced?

My instinct is that storing them separately is cleaner, as long as we could perform the same slicing so that the dimensions match.

samaloney:
@jajmitchell 
> My instinct is that storing them separately is cleaner, as long as we could perform the same slicing so that the dimensions match.

Yea I think this is the best approach we maybe able to incorporate them in the meta data later and use its slicing machinery.

KriSun95 (approved):

jajmitchell (approved):

PR #250, opened 2026-02-03
Deprecrate legacy and seperate example galleries

samaloney:
## PR Description

* Deprecate legacy, add warning on module import and add warnings call outs in the documentation.
* Split gallery into Example and Legacy sections


docs/how_to/index.rst
```diff
@@ -11,3 +11,4 @@ you quickly accomplish common tasks.
    :maxdepth: 1
 
    ../generated/gallery/index
+   ../generated/gallery/legacy/index
```
KriSun95, line 18: Building the docs locally, it looks like this builds them such that the links in the menu at the side are normal (one for the example gallery and one for the legacy gallery) but the links on the page show two for the legacy gallery.

I think I get what's happening but playing about with it, I found this fixed it but I'm not sure if this is the "correct" way or just a hack:
```
.. toctree::
   :maxdepth: 1

   ../generated/gallery/index
.. toctree::
   :maxdepth: 1
   :hidden:

   ../generated/gallery/legacy/index
```
Now my local docs show one link for each in the side menu AND one each on the page.

KriSun95 (changes requested):
Other than the comment about multiple links it LGTM!

Happy to approve if the weird link issue is only on my machine too.

docs/how_to/index.rst
jajmitchell: I also experience this issue. The docs build also fails for me locally too, I think its the sphinx issue again

samaloney: ~~Are you building with `tox -re build_docs` after doing an `rm -r _build generate api`? As the CI is passing so could be local issues?~~

Never mind I see on the the RTD build https://sunkit-spex--250.org.readthedocs.build/en/250/how_to/index.html

samaloney, line 18:
```suggestion

```

samaloney:
Ok it's sorted now https://sunkit-spex--250.org.readthedocs.build/en/250/how_to/index.html

KriSun95:
I didn't realise we could access the docs build from the tests in the PR 🤯

The sidebar menu now only shows the link to the Example Gallery. I got that behaviour when I just removed the link to `../generated/gallery/legacy/index` in the `docs/how_to/index.rst` file instead of doing the `:hidden:` thing I suggested. 

I suppose my question would be: do we want the behaviour that the legacy gallery is slightly more hidden? I could be convinced either way to be fair.

samaloney:
Yea it's because the legacy gallery is a subgallery of the example gallery. I didn't want to have to create a 2nd subgallery as I'd have to name it `Current` or something. So I've used the approach @KriSun95 suggested for the moment.

KriSun95 (approved):
All makes sense! LGTM

jajmitchell (approved):

PR #262, opened 2026-04-16
Add caching of CHIANTI files to GHA 

samaloney:
## PR Description

Use openastronomy GHA templates caching features to cache the CHIANTI files.

So what should ideally happen with this PR is that once it's merge to main it will run and generate the cache entries for each os with keys ending in `refs/heads/main` e.g. `sunpy-data-linux-refs/heads/main` then any subsequent run will restore the cache from these keys

## AI Assistance Disclosure

AI tools were used for:
- [ ] Code generation (e.g., when writing an implementation or fixing a bug)
- [ ] Test/benchmark generation
- [ ] Documentation (including examples)
- [ ] Research and understanding
- [x] No AI tools were used

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -36,10 +36,9 @@ jobs:
       submodules: false
       coverage: codecov
       toxdeps: tox-pypi-filter
-      cache-path: |
-        $HOME/sunpy/data_manager
-      cache-key: sunpy-data-linux-${{ github.ref }}
-      cache-restore-keys: sunpy-data-linux-refs/heads/main
+      cache-path: ~/sunpy/data_manager
+      cache-key: sunpy-data-${{ runner.os }}-${{ github.ref }}
+      cache-restore-keys: sunpy-data-${{ runner.os }}-refs/heads/main
       envs: |
         - linux: py313
     secrets:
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -36,6 +36,10 @@ jobs:
       submodules: false
       coverage: codecov
       toxdeps: tox-pypi-filter
+      cache-path: |
+        ~/sunpy/data_manager/*.sav
+        ~/sunpy/data_manager/*.genx
```
samaloney: Just cache the entire directory `~/sunpy/data_manager/`?

samaloney:
Since these should not change maybe use broader cache key `sunny-data-<os>` that way once cached should always be there - or maybe on update to main `sunny-data-main-<os>` that way they are updated once in while?

Can also maybe cache the same data for the doc builds?

.github/workflows/ci.yml
```diff
@@ -62,13 +66,31 @@ jobs:
...
       coverage: codecov
       toxdeps: tox-pypi-filter
+      cache-path: |
+        $HOME/sunpy/data_manager
+      cache-key: sunpy-data-${{ github.ref }}
       posargs:
       envs: |
         - linux: py314
+          cache-path: $HOME/sunpy/data_manager
+          cache-key: sunpy-data-linux-${{ github.ref }}
```
Cadair, line 72: Is there not a github context var for OS name?

samaloney: Maybe I could use [`runner.os`](https://docs.github.com/en/actions/reference/workflows-and-actions/contexts#runner-context) and have single cache block for the all the envs

samaloney: Nope doesn't have access to the runner info at that stage

Cadair (approved):
Seems reasonable to me, need to merge it and experiment on future PRs to check exactly how it works.

PR #286, opened 2026-07-15
Fitter albedo fix

### LINKED ISSUE #284: Albedo issue in legacy fitting?
There appears to be some odd behaviour in the legacy fitting/plotting when using albedo correction and RHESSI data (it might affect other data as well). The albedo component appears to be larger than expected as well as adding bumps (often when the energy bin widths change). The previous discussion of albedo was in #128 but that was mostly about the implementation

The code and data used to produce these plots are available in this repo [ianan/albedo_test](https://github.com/ianan/albedo_test/tree/main) and the issue is with the grey albedo component, and its contribution to the total model, i.e.

Image

In comparison ospex gives you

Image

To me there appears to be two issues here:

- Something odd in the fitting/plotting in the legacy version
- Something related to interpolation/bin widths

I've reached that conclusion because if I calculate the albedo component in count space $C_A$ and compare it to what I can determine from the output of `fitter.plotting_info` the results are quite different, i.e.

Image

Here I have calculated $C_A$ three different ways which correspond to how the legacy code should be doing it (directly fold through, pink) , the new plotter does it (photon spec addition difference, blue) and how ospex does it (modified response difference, black). See [ianan/albedo_test/README.md](https://github.com/ianan/albedo_test/blob/main/README.md) for more details maths on each of these.

Reassuringly they all give the same answer (which they mathematically should) just what the legacy fitter is actually doing is different.

These manually calculated ones are closer to what OSPEX gives, however they still contain smaller bumps, some of which are where the energy bin widths change - so this implies some sort of interpolation/bin width issue somewhere as the albedo green matrix are given with 1 keV binning.

Another example using 1 keV binned RHESSI data for a different flare still has fewer issues in `fitter.plotting_info` but the bumps disappear completely in my manually calculated ones (so no/minmal interpolation)

Image

Next steps:

- This can't just be an issue I'm having with RHESSI data - can someone else replicate it from  [ianan/albedo_test](https://github.com/ianan/albedo_test/tree/main) ?
- Someone check this data/fit with the new sunkit-spex fitting/plotting. If it matches OSPEX then great, if only minor bumps then it's "just" an interpolation issue and something the bigger issue is with the legacy approach.
- Is the issue with the legacy approach just in plotting or also in fitting? Does it produce the same issue for STIX data? 
- What is going on with the interpolation of the albedo green matrix?

ianan:
## PR Description

Fixes #284 by changing the order in which the photon bin widths are applied when including the albedo correction to the photon model and folding it through the SRM to produce a count model. Instead of the photon spectrum being multiplied by the photon bin widths before being folded through the albedo matrix, this now happens to the SRM before being used to convert the photon spectra to counts. 

Change implemented by adding photon bin widths to `make_model`, and if they are included new way of calculating used. All other inputs/output to `make_model` remain the same.

Also 
- made the albedo component clearer in default plots (now a labelled plum coloured line).
- the albedo count spectrum is also stored/logged in plotter_info for separate plotting via `fitter.plotting_info['spectrum1']['albedo_excess_count']`.
- modified formatting of fit parameters in default plot to better match their scale (needed here or just a nice bonus?)

## AI Assistance Disclosure

AI tools were used for:
- [x] Code generation (e.g., when writing an implementation or fixing a bug)

### DIFF
--- a/changelog/286.bugfix.rst
+++ b/changelog/286.bugfix.rst
@@ -1,0 +1,1 @@
+Fixed a bug in legacy albedo correction that causes bumps at energy bin width changes, by applying photon bin width scaling to SRM instead of albedo matrix in :func:`sunkit_spex.legacy.fitting.fitter.make_model`.
--- a/sunkit_spex/legacy/fitting/fitter.py
+++ b/sunkit_spex/legacy/fitting/fitter.py
@@ -1474,14 +1474,21 @@ class Fitter:
             # assign the spectrum parameter values to their model param counterpart
             sep_params = dict(zip(self._orig_params, ordered_kwarg_values))
 
-            # calculate the [photon s^-1 cm^-2]
+            # calculate the [photon s^-1 cm^-2] -> keep this as before
             m = (
                 self._model(**sep_params, energies=kwargs["photon_channels"][s]) * kwargs["photon_channel_widths"][s]
             )  # np.diff(kwargs["photon_channels"][s]).flatten() # remove energy bin dependence
 
             # fold the photon model through the SRM to create the count rate model, [photon s^-1 cm^-2] * [count photon^-1 cm^2] = [count s^-1]
             cts_model, cts_albedo_exess = make_model(
-                energies=kwargs["photon_channels"][s], photon_model=m, parameters=None, srm=kwargs["total_responses"][s], albedo_corr=self.albedo_corr, albedo_angle =self.albedo_angle, albedo_anisotropy=self.albedo_anisotropy
+                energies=kwargs["photon_channels"][s],
+                photon_model=m,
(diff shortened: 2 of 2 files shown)


jajmitchell (approved):
LGTM

samaloney (commented):
Eh looks good think just needs to be rebased with main to get the doc build to work.

ianan:
> Eh looks good think just needs to be rebased with main to get the doc build to work.

Is that something I need to do?

samaloney:
@ianan I've rebase this on main so the doc should hopefully build now

KriSun95 (approved):
LGTM too!

samaloney (approved):

PR #293, opened 2026-07-30
Add warm thick

KriSun95:
Opening this WIP PR that introduces the warm thick target model...at least the core part of it

The warm thick target is the addition of the thick target and a thermal component; that thermal component is the ``ThermalEmission`` model where the emission measure input is calculated from the thick target's low-energy cut-off and electron flux values, and new plasma number density and column length parameters.

I've made a new model called ``ThickTargetWarmContribution`` that, given these parameters, will calculate the thermal component o the whole warm thick target model we know rom IDL. That should mean that to re create the whole warm thick target model we would only need to do:

```
warm_thick = ThickTarget()+ThickTargetWarmContribution()
```
and make sure the low-energy cut-off and electron flux parameters are shared.

It is a work in progress, still need to add tests and comparisons to IDL, etc. but we'll want this physical model in eventually I'd imagine so might as well put it here for folk to see.

Here is a snippet of code I've been using it in:

```python
import astropy.units as u
import matplotlib.pyplot as plt
import numpy as np

from sunkit_spex.models.physical.nonthermal import ThickTarget
from sunkit_spex.models.physical.thermal import ThermalEmission,  ThickTargetWarmContribution

energy_edges = np.arange(1.6, 15, 0.1) << u.keV

energy_mids = (energy_edges[:-1]+energy_edges[1:])/2

## parameters from https://iopscience.iop.org/article/10.3847/2041-8213/ab7341 Table 1 Lindsay's paper
temperature = 10.2 << u.MK
plasma_density = 6e9 << u.cm**-3
low_e_cutoff = 6.5 << u.keV
total_eflux = 1.8e35 << (u.electron/u.second) 
p = 6.3
length = 15 << u.Mm

model_thick = ThickTarget(total_eflux=total_eflux/1e35, p=p, low_e_cutoff=low_e_cutoff, break_energy=1500<< u.keV, q=20, high_e_cutoff=1500<< u.keV)
model_warmtherm2 =  ThickTargetWarmContribution(temperature=temperature,
                                                plasma_density=plasma_density,
                                                low_e_cutoff=low_e_cutoff,
                                                total_eflux=total_eflux,
                                                length=length,
                                                )
# print(model_warmtherm2.parameters)
flux_c2 = model_warmtherm2(energy_edges)
# print(model_warmtherm2.parameters)
flux_e2 = model_warmtherm2.evaluate(energy_edges, *model_warmtherm2.parameters)

plt.figure()
[...]

PR #291, opened 2026-07-28
Thick Target (and thermal) Update

KriSun95:
This PR got a bit away from me but it was all in the investigation of trying to figure out why the Thick target model would return NaNs sometimes.

So, what I've done is:

- Because the scaling of emission measure and electron fluxes was being a real pain in the neck trying to work out what was going on, I've defined new Astropy units for these parameters to work in. This means that our units can now actually make sense using ``norm_thick_target_eflux_units``, ``norm_thin_target_eflux_units``, and ``norm_thermal_emission_measure_units``.
  - E.g., we can pass the real value of emission measure, say ``5e49 cm^-3``, if we like but internally the value will be converted to the scaled version of ``5 new_units`` where ``new_untis`` is ``1e49 cm^-3``. This is unit aware and so conversions can easily happen that the code knows about rather than it being in some places in the doc string that ``5 cm^-3`` really means ``5e49 cm^-3``.
  - Tests were added tp make sure ``5e49 cm^-3`` and ``5 new_units`` work the same and as intended.
- I've added realistic bounds to the thick target and thermal model parameters. Particularly for the thick target, certain values would result in a division by zero somewhere in the internal calculations and sometimes cause NaNs.
  - Tests have been added to make sure these weird values that the new bounds try to prevent give ``RunTimeWarnings`` as they should.
- Units for parameters are now assumed or converted with ``<<`` when the model evaluates. This means we don't need to have ``if``/``else`` statements checking all the input units then assuming the units if they're not there.
  - Unit behaviour should now be consistent between the models. Before, the thermal and thick target would work differently. E.g., the thermal model would always return units when evaluated but the thick target wouldn't. Now both return them, always in the same way (@jajmitchell , would this affect fitting? I'm unsure how if the thermal and thick were behaving differently anyway but want to check).

The above were fixes I made to keep investigating the weird NaN behaviour I've seen before, removing cases where this has happened. I then ran a fit, producing the plot below (data=thermal+thick source, model=thermal+thick model, thick=just the thick target model contribution):

Screenshot 2026-07-28 at 3 45 46 pm

From here, the thick target is returning one bin around 20 keV with a NaN. 

Investigating this further, I think there are two important points:

- **it's the bin with the low-energy cut-off energy in it**. See the example code and output below.
- **Moreover, this only appears with finer energy binning. Running the Thick target model with energies like ``energy_edges = np.arange(2, 15, 0.1) << u.keV`` instead does not return a NaN in that bin.**

Still to do:

- [x] create a changelog
[...]

PR #295, opened 2026-08-04
Non thermal models clean up and refactor

samaloney:
## PR Description

Less code, more readable code, and some performance improvements

* Remove some unused and unnecessary thin wrappers
* The old `gauss_legendre` port and the `fixed_quad_batch` are almost identical which should now be clearer the later is numerically better as the scaling operations are not compounded in the integration as they are done at the end - changed default to `fixed_quad_batch`
* Figure out source of some of the numeric constants and added comment with units
* Pull out a bunch of static constants to module level vars 
* Modernised some of the code structure and eliminated unnecessary checks
* More / improved comments
* More edge case tests
* ThickTarget ~ 1.5x and ThinTarget ~2x faster

- [ ] Still importing constants from legacy should we just switch to the astropy values at this stage or also pull out the constant function from legacy

## AI Assistance Disclosure

AI tools were used for:
- [  ] Code generation (e.g., when writing an implementation or fixing a bug)
- [x ] Test/benchmark generation
- [  ] Documentation (including examples)
- [x ] Research and understanding
- [  ] No AI tools were used

### DIFF
--- a/sunkit_spex/models/physical/integrate.py
+++ b/sunkit_spex/models/physical/integrate.py
@@ -138,8 +138,8 @@ def fixed_quad_batch(func, a, b, n=5, args=(), func_kwargs={}):
     1.0
 
     """
-    a = np.array(a)
-    b = np.array(b)
+    a = np.asarray(a)
+    b = np.asarray(b)
     standard_nodes, standard_weights = _cached_roots_legendre(n)
     nodes = (b - a).reshape(-1, 1) * (standard_nodes + 1) / 2.0 + a.reshape(-1, 1)
     return np.squeeze(
--- a/sunkit_spex/models/physical/integrate.py
+++ b/sunkit_spex/models/physical/integrate.py
@@ -11,44 +11,9 @@ def _cached_roots_legendre(n):
     """
     Cache the Gauss-Legendre nodes and weights for a given order.
 
-    `scipy.special.roots_legendre` is not itself cached, and this function is called on every
-    call to `gauss_legendre`/`fixed_quad`, so this avoids recomputing the same nodes and weights
-    repeatedly during the doubling-order integration in
-    `~sunkit_spex.models.physical.nonthermal._integrate_part`.
+    Similar to the caching used internally in `~scipy.integrate.fixed_qaud` but using the public API
     """
     return roots_legendre(n)
-
-
-def _legendre_roots(a, b, n=5):
-    """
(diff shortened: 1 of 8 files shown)


sunkit_spex/models/physical/integrate.py
samaloney, line 1: It much easier to see that `gauss_legendre` and ``gauss_legendre`` do almost the same thing the only difference is when the mapping to the real limits is done before `gauss_legendre` or after in `gauss_legendre` which mean the numerical error aren't folded in to the integration

samaloney:
Just some notes I think this code is now close the limit of the possible performance without going down the Cython or numba routes which have there own complications.

The main bottle necks are `bremsstrahlung_cross_section` and `BrokenPowerLawElectronDistribution.density` which take up ~50% of the time.

Figure shows a performance comparison between main, this branch and SSWIDL function. The main reason `think_target` is faster is that it doesn't require the `.density` method.
image

sunkit_spex/models/physical/nonthermal.py
```diff
@@ -1,33 +1,70 @@
...
 import astropy.units as u
 from astropy.modeling import FittableModel, Parameter
 
-from sunkit_spex.legacy import constants as const
-from sunkit_spex.legacy.integrate import gauss_legendre
+from sunkit_spex.legacy import constants
+from sunkit_spex.models.physical.integrate import fixed_quad_batch
 
-const = const.Constants()
+const = constants.Constants()
```
KriSun95, line 25: I like the idea of having our own specific constants (e.g., from hard-coded IDL sources and now just Astropy ones). One thought I had is that it would make sense in the future to add Astropy units to these. I realised this was useful in #293

sunkit_spex/models/physical/nonthermal.py
```diff
@@ -119,29 +148,20 @@ def __init__(
...
+        output = bremsstrahlung_thick_target(
+            (energy_centers << u.keV).value,
+            p,
+            (break_energy << u.keV).value,
+            q,
+            (low_e_cutoff << u.keV).value,
+            (high_e_cutoff << u.keV).value,
+            self.integrator,
+        )
+        output[~np.isfinite(output)] = 0
```
KriSun95, line 162: This can be addressed in another PR but I think I included this to avoid weird values in the output. In #291 I think I remove this which forced me to try and track down when things weren't quite right.

KriSun95 (approved):

jajmitchell (approved):

PR #301, opened 2026-08-09
Type checking

samaloney:
## PR Description

- [x] Add mypy and pyright type check configuration and CI job
- [x] Update code to pass new CI type check job
- [ ] Review narrow/widen some types for Quantities etc just got it pass right now

If we're want to this it would be good to agree and merge sooner rather than later as touches nearly every file. 

If we don't want the types inline could move them to stubs files

## AI Assistance Disclosure

Help figure out some of the types and specific inline ignores and configuration 

AI tools were used for:
- [ ] Code generation (e.g., when writing an implementation or fixing a bug)
- [ ] Test/benchmark generation
- [ ] Documentation (including examples)
- [X ] Research and understanding
- [ ] No AI tools were used

### DIFF
--- a/sunkit_spex/fitting/objective_functions/optimising_functions.py
+++ b/sunkit_spex/fitting/objective_functions/optimising_functions.py
@@ -2,10 +2,22 @@
 This module contains functions that can evaluate models and return a fit statistic.
 """
 
+from typing import Any
+from collections.abc import Callable
+
+import numpy as np
+from numpy.typing import NDArray
+
 __all__ = ["minimize_func"]
 
 
-def minimize_func(params, data_y, model_x, model_func, statistic_func):
+def minimize_func(
+    params: NDArray[np.float64],
+    data_y: NDArray[np.float64],
+    model_x: NDArray[np.float64],
+    model_func: Any,
+    statistic_func: Callable[[NDArray[np.float64], NDArray[np.float64]], float],
+) -> float:
     """
     Minimization function.
 
--- a/sunkit_spex/models/physical/nonthermal.py
+++ b/sunkit_spex/models/physical/nonthermal.py
@@ -1,17 +1,24 @@
 import logging
+from typing import Any
 
 import numpy as np
+from numpy.typing import NDArray
 
 import astropy.units as u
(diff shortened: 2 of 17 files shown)


sunkit_spex/fitting/objective_functions/optimising_functions.py
```diff
@@ -2,10 +2,22 @@
...
 __all__ = ["minimize_func"]
 
 
-def minimize_func(params, data_y, model_x, model_func, statistic_func):
+def minimize_func(
+    params: NDArray[np.float64],
+    data_y: NDArray[np.float64],
+    model_x: NDArray[np.float64],
+    model_func: Any,
+    statistic_func: Callable[[NDArray[np.float64], NDArray[np.float64]], float],
```
KriSun95, line 19: How does this work if the Callable can have keyword arguments other varying inputs?

sunkit_spex/models/physical/nonthermal.py
```diff
@@ -95,15 +102,15 @@ class ThickTarget(FittableModel):
...
-    ):
+        p: Any = p.default,
+        break_energy: Any = u.Quantity(break_energy.default, break_energy.unit),  # pyright: ignore[reportArgumentType]
+        q: Any = q.default,
+        low_e_cutoff: Any = u.Quantity(low_e_cutoff.default, low_e_cutoff.unit),  # pyright: ignore[reportArgumentType]
+        high_e_cutoff: Any = u.Quantity(high_e_cutoff.default, high_e_cutoff.unit),  # pyright: ignore[reportArgumentType]
+        total_eflux: Any = u.Quantity(total_eflux.default, total_eflux.unit),  # pyright: ignore[reportArgumentType]
+        integrator: Any = None,
+        **kwargs: Any,
+    ) -> None:
```
KriSun95, line 113: I think all of these in ``__init__`` should be Astropy quantities even if dimensionless.

KriSun95 (approved):
Adding types to be explicit is a great idea. I'd imagine this is something we'll continue to edit and adopt as we keep building too.

jajmitchell (approved):

Issue #132, opened 2023-11-28
RHESSI fits sum_flag=0 breaks code


ianan:
### Describe the bug

At the moment the code returns an error and breaks when trying to load a RHESSI srm/spec fits file with sum_flag != 1 in the header.  I guess @KriSun95 did this is to avoid issues of multiple detectors that aren't summed, i.e. expecting always to have a single srm and single spectrogram in each fits file. But even if there is a single srm/spec in there (say only using a single RHESSI detector) will get the error/break if sum_flag=0 in the fits header.

Short term solution is just to make sure sum_flag=1 when generating RHESSI srm/spec fits from the hsi_spectrum() in sswidl, whether summed over multiple detectors or "summed" over a single detector.

Longer term might want a smarter way to deal with this? Maybe @settwi might have some ideas given your work on loading in RHESSI spec/srm.

### To Reproduce

Load in a RHESSI spec/srm fits file where have a single detector spectrogram/srm but sum_flag=0 in the header, i.e. https://github.com/ianan/rhessi_spectra/blob/main/mfstats9_fits/20030725_082642_srm_org_d1.fits https://github.com/ianan/rhessi_spectra/blob/main/mfstats9_fits/20030725_082642_spec_org_d1.fits

`from sunkit_spex.sunxspex_fitting.fitter import Fitter`
`spec = Fitter(pha_file="20030725_082642_spec_org_d1.fits", srm_file="20030725_082642_srm_org_d1")`

Gives error/breaks in sunxspex_fitting/instruments.py:662 with warning:
"Apparently spectrum file's SUMFLAG should be one and I don't know what to do otherwise at the moment."

settwi:
I believe OSPEX only allows spectroscopy to be performed when `SUMFLAG=1` (see [here](https://hesperia.gsfc.nasa.gov/ssw/packages/spex/idl/object_spex/spex_hessi_fits2drm.pro)):
```idl
sum_flag = fxpar (hdr, 'SUMFLAG', count=count)
if count gt 0 and sum_flag eq 0 then begin
  if n_elements(str2arr(detused, ' ')) gt 1 then begin   ; check for > 1 det
	  err_code = 1
	  err_msg = 'OSPEX can not handle SRM files with >1 detectors that are not summed (sum_flag=0). Aborting'
	  return
	endif
endif
```

Summing counts from the different RHESSI detectors is kind of fraught so that makes sense.
[...]

PR #304, opened 2026-08-12
RHESSI: allow SUMFLAG=0 for single-detector files

settwi:
Partially addresses #132. I would argue it closes that issue.


settwi:
@ianan please give this a shot. it is a simple change

